PR #1, opened 2022-12-16
Start of API pull workflow

steeleb:
Didn't get terribly far (aka, this is not functional yet), just trying to trouble shoot how to set up virtual environments so that we can use the Python API for HydroVu.

Set up backbone of credentialing via yml and readme. (97ca408, 8e09231)
Worked on creating a *reproducible* conda env, need to give up for now (e50c171)

I'm not going to work more on this until next week, and if you don't get to this before your time off, don't worry about it. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ docs/
 backup_data/
 backup_docs/
 misc_playtime/
+*/credentials.yml
+*credentials.yml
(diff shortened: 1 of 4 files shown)


kathryn-willi (approved):
This is great initial progress, thank you so much for your help!

PR #4, opened 2022-12-30
Added timelapse/wq code

kathryn-willi:
I added code to overlay water quality data with time lapse footage, and tidied up the code that QA/QC's the raw data. Developed functions to pull in outside data (e.g., FC storm data, LC diversion data).


scratch.Rmd
```diff
@@ -0,0 +1,556 @@
...
+                      Turbidity_NTU = contains('Turbidity'),
+                      #TSS_mg_L = contains('Total.Suspended.Solids..mg.L.'),
+                      #Chla_RFU = contains('Chl.a.Fluorescence..RFU.'),
+                      #Chla_µg_L = contains('Chl.a.Concentration..µg.L.'),
+                      #Pressure_PSI = contains('Pressure..psi.'),
+                      Depth_ft = contains('Depth..ft')) %>%
+                      #Elevation_m = contains('Level..Elevation..m.'),
+                      #Depth_to_Water_ft = contains('Level..Depth.to.Water..ft.')) %>%
+  mutate(DT = ymd_hms(DT)) %>%
+  mutate(DT = DT - lubridate::hours(1)) %>% # reported in MDT >:(
```
steeleb: Note that you can store the `ymd_hms` with a tz argument, then use `with_tz()` to convert datetime stamps to another timezone. This is especially helpful when you're dealing with data that have been converted on DST beyond DST start/end or when you do happen to have data outside of DST that need to be managed as UTC datetimes so you don't eliminate data with overlapping time stamps in the fall change. This is the most easily-read list of tz's: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones; where you'd use 'TZ data base name' as your tz argument in `ymd_hms()`

scratch.Rmd
```diff
@@ -0,0 +1,556 @@
...
+                      Depth_ft = contains('Depth..ft')) %>%
+                      #Elevation_m = contains('Level..Elevation..m.'),
+                      #Depth_to_Water_ft = contains('Level..Depth.to.Water..ft.')) %>%
+  mutate(DT = ymd_hms(DT)) %>%
+  mutate(DT = DT - lubridate::hours(7)) %>%
+  rbind(rawless_troll_legacy) %>%
+  mutate_at(vars(2:ncol(.)), as.numeric) %>%
+  mutate(site = "legacy") %>%
+  arrange(ymd_hms(DT)) %>%
+  mutate(DT = as.character(round_date(ymd_hms(DT), "15 minutes"))) %>%
```
steeleb: Highly suggest not overwriting the DT column to this rounded time. store in a DT_corr, or conversley (and maybe more straightforward, save original DT in DT_instrument column so that you retain the original data.

scratch.Rmd
```diff
@@ -0,0 +1,556 @@
...
+  mutate(DT = ymd_hms(DT)) %>%
+  full_join(raw_tube_legacy, by='DT') %>%
+  dplyr::bind_rows(., filter(pre_2022, site == "legacy")) %>%
+  full_join(filter(field_notes, site == "legacy"), by = c('DT','site')) %>%
+  mutate(date = as_date((DT)),
+         hour = hour(DT),
+         year = year(DT)) %>%
+  arrange(ymd_hms(DT)) %>%
+   # remove data ranges where sensor was pulled out of field (THAT I KNOW OF!).
+  dplyr::filter(!ymd_hms(DT) >= ymd_hms('2021-12-04 19:30:00') & ymd_hms(DT) < ymd_hms('2022-04-06 17:30:00'),
```
steeleb: instead of filter, consider using a function (over your list of dates where you have visits listed) where you recode data to NA instead of dropping observations. An example where I use a loop is here (https://github.com/Lake-Sunapee-Protective-Association/LSPA_buoy/blob/master/Sunapee_buoy2_2021.R) starting at line 182.

scratch.Rmd
```diff
@@ -0,0 +1,556 @@
...
+dygraph(timeseries, main="Depth Timeseries (UTC)") %>% 
+ dyRangeSelector()
+```
+
+## Legacy conductivity
+```{r}
+legacy_conductivity <- dplyr::select(rawless_legacy, c(DT, Specific_Conductivity_µS_cm, site)) %>%
+  arrange(ymd_hms(DT)) %>%
+  group_by(site,DT) %>%
+  summarize(Specific_Conductivity_µS_cm = mean((Specific_Conductivity_µS_cm))) %>%
```
steeleb: is this needed because of dupe rows? Before you summarize, consider why it is needed and if there is a way you can fix it in your data instead of summarizing. If you *do* need to summarize in this way, your variable is no longer just Spec Cond, it's an average Spec Cond (changing the col name), and it would be helpful for an end user to know how many values contribute to that mean value (`n_SpecCond = length(SpecCond)`), and what the standard deviation was (`SD_SpecCond = sd(SpecCond)`).  IMO, you're creating a 'Level 2' dataset where you are augmenting the observations. While this may be necessary for your uses, another user might not agree.

tidying_2022.Rmd
```diff
@@ -45,17 +67,28 @@ library(zoo)
...
-This is the first year that there are consistent field notes. 
+#### Pulling in previous years' data
 
 ```{r}
-# WHEN DATA WAS STORED ON GOOGLE DRIVE:
-# field_notes <- read_sheet("https://docs.google.com/spreadsheets/d/11Gc1eS1wt9NrU12hWw1mUl25ixKZrTndIVCsvD4UKao/edit#gid=0") %>%
-#   mutate(start_time_mst = (substr(start_time_mst, 12, 19)),
-#          DT = round_date(as_datetime(paste0(date,' ',start_time_mst), tz="MST"), '15 mins'))
+pre_2022 <- read_csv('data/pretty/all_sensor_data_pre2022.csv') %>%
+  mutate(site = case_when(sensor == "Rist" ~ "rist",
```
steeleb: you could also use:
'mutate(site = to_lower(site))'

steeleb (approved):
See notes in scratch.Rmd - I looked at that one first and made notes through the top half of the document.

steeleb:
Lots of progress here! Well done. I made some comments in the scratch code (mostly to help with DT and functions), mostly because I didn't know where to start or where to focus. Looking further in your commit history, I probably could have workd from the most recent backwards. Sorry if some of the comments are unuseful.

Make sure in your PRs you name a focus area and what you want your reviewer to look at. In general, this is a TON of code and honestly, too much to review in one sitting without direction. Try to make PRs more frequently so your iCP only has to review ~ 500 lines of code. That's a SINGLE file in this PR. :). B

PR #11, opened 2023-03-09
more conservative dry filtering

kathryn-willi:
For Rist, added more filtering steps to remove points where the sensor was likely not submerged. Advise we also send Jared Depth_ft along with SC and Temp!


SamStruthers (commented):
Loooks great

PR #16, opened 2023-08-10
Rebeginning QAQC workflow

kathryn-willi:
I've started working on initial flagging pipelines, namely for temperature at a single site for now. If you could please review what I've put together in the 01-organize_raw_data.Rmd file, that would be amazeee. In particular, the final code chunk where I develop the data flags.


steeleb (approved):
WE'RE DOING A THING!! 

I've got edits, but instead of typical code review, we're gonna do the pair programming, because we're a bad ass team. I'm going to merge this and then send you a PR with my updates.

B

Issue #19, opened 2023-08-17
Ensuring site visits (and therefore times when sonde was out of the water) are represented as NA's in the Level 1 data set.


kathryn-willi:
For Level 1 cleaning, any data that is recorded when the sonde is out of water should be re-coded to NA (i.e. site visits), with the OG data kept only in Level 0 datasets. This flagging step should also happen across all data (aka, earlier in the workflow) so that you don't have to run this step for every site and every parameter.

PR #26, opened 2023-08-18
Adding summarization function for data that was pulled from HydroVu API

juandlt-csu:
This PR does not address any issues in this repo. This PR creates a function that is used to map over each combination of site and parameter that is pulled from the HydroVu API to summarize the information and begin flagging it for site visits and missing data.

Please review that this function and downstream flagging is working properly. I think that the creation and the results of the `where` column should still be inspected carefully.

### DIFF
--- a/src/qaqc/01-organize_raw_data.Rmd
+++ b/src/qaqc/01-organize_raw_data.Rmd
@@ -156,12 +156,12 @@ field_notes <- read_excel("data/sensor_field_notes.xlsx") %>%
          site = tolower(site),
          season = year(DT_round)) %>%
   arrange(DT_round) %>%
-  # binary version of deployed/pulled options. may need to play around with this som
+  # binary version of deployed/pulled options. may need to play around with this some
   # more
   mutate(where = case_when(!is.na(sensor_pulled) & !is.na(sensor_deployed) ~ 0,
                            !is.na(sensor_pulled) & is.na(sensor_deployed) ~ 1,
                            is.na(sensor_pulled) & !is.na(sensor_deployed) ~ 0,
-                           is.na(sensor_pulled) & is.na(sensor_deployed) ~ 0))
+                           is.na(sensor_pulled) & is.na(sensor_deployed) ~ 0)) 
 ```
 
 Munge field notes
@@ -260,6 +260,8 @@ processing speed (targets?). Here I am starting a pipeline for Archery's tempera
 #### Format data
 
 ```{r}
+# This chunk is giving me issues...
+# What I want to happen is to keep the flag comment as a site visit from all_data
 temperature_archery <- all_data %>%
   filter(parameter == "Temperature" & site == "archery") %>%
   select(-name) %>%
@@ -267,21 +269,21 @@ temperature_archery <- all_data %>%
   group_by(DT_round) %>%
   # I WOULD LIKE TO PRESERVE THE RAW VALS THAT WENT INTO MAKING THIS AVERAGE
   # IN A NESTED DATAFRAME VIA NEST(). IT WON'T WORK >:(
-  summarize(p1 = as.numeric(mean(value, na.rm = T)),
+  summarize(p1 = as.numeric(mean(value, na.rm = T)), 
             diff = abs(min(value, na.rm = T) - max(value, na.rm = T)),
             n_obs = n()) %>% 
   ungroup() %>%
-    arrange(DT_round) %>%
+  arrange(DT_round) %>%
   # pad the dataset so that all 15-min timestamps are present
   pad(by = "DT_round", interval = "15 min") %>%
   mutate(DT_join = as.character(DT_round)) %>%
   full_join(filter(dplyr::select(field_notes, where, visit_comments, sensor_malfunction, DT_join, site), site == "archery"), 
-            by = c('DT_join'))
+            by = c('DT_join')) # adding from DT_join here adds things that are not necessarily from temperature data
 
 temperature_archery <- cbind(select(temperature_archery, -where), data.table::setnafill(x = temperature_archery[,6], type = "locf")) %>%
   # add flag for no data times
   mutate(flag = if_else(is.na(p1) | where == 1, 'no data', NA_character_)) %>%
-  dplyr::filter(is.na(flag)) %>%
+  dplyr::filter(is.na(flag)) %>% # what does this do?
   pad(by = "DT_round", interval = "15 min") %>%
   # ... so that we can get the proper leading/lagging values across our entire timeseries:
   mutate(front1 = lead(p1, n = 1),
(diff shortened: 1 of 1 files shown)


steeleb:
I'll take a peek at this first thing Monday! This PR request is super clear - thank you!

Looking forward to reviewing this!

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -260,28 +260,30 @@ processing speed (targets?). Here I am starting a pipeline for Archery's tempera
...
   mutate(DT_join = as.character(DT_round)) %>%
   full_join(filter(dplyr::select(field_notes, where, visit_comments, sensor_malfunction, DT_join, site), site == "archery"), 
-            by = c('DT_join'))
+            by = c('DT_join')) # adding from DT_join here adds things that are not necessarily from temperature data
 
 temperature_archery <- cbind(select(temperature_archery, -where), data.table::setnafill(x = temperature_archery[,6], type = "locf")) %>%
   # add flag for no data times
   mutate(flag = if_else(is.na(p1) | where == 1, 'no data', NA_character_)) %>%
-  dplyr::filter(is.na(flag)) %>%
+  dplyr::filter(is.na(flag)) %>% # what does this do?
```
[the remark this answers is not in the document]
kathryn-willi: ^^ like: `filter( !is.na(p1) | where != 1) %>% pad(by = "DT_round", interval = "15 min")`

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -301,6 +303,70 @@ ggplot(temperature_archery, aes(x = DT_round, y = p1)) +
...
+  
+  print(paste0('Trying site ', site_arg, " with parameter ", parameter_arg, "."))
+  
+  # filtering the data and generating results
+  summary <- tryCatch({ 
+    api_data %>%
+    filter(site == site_arg & parameter == parameter_arg) %>% 
+    select(-name) %>%
+    distinct() %>%
+    group_by(DT_round, site, parameter) %>%
```
kathryn-willi, line 324: somewhere in here, I suggest adding in that filtering step discussed in previous comment.

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -314,6 +380,16 @@ slope_thresh = 1/15
 
 #### Add the flags
 
+```{r}
+# Add flag column to all_data_summary_df and begin flagging data
```
kathryn-willi, line 384: suggest integrating this flagging in the function since this won't change across site/parameter types

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -322,14 +398,24 @@ temperature_archery <- temperature_archery %>%
   mutate(flag = if_else(p1 >= temp_max | p1 <= temp_min, "Out of bounds", flag),
          p1 = if_else(p1 >= temp_max | p1 <= temp_min, NA_real_, p1))
 
+temperature_archery <- temperature_archery %>%
```
kathryn-willi, line 401: Out of range, slope, sd stuff WILL be different across site/parameter types so good job keeping this out of the upstream function :-)

kathryn-willi (approved):
Great first PR! I really like how you have commented throughout... It really helped direct me to locations in the workflow you wanted help/needed some feedback!

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -301,6 +303,70 @@ ggplot(temperature_archery, aes(x = DT_round, y = p1)) +
...
 
+```{r}
+# Determine each site and parameter in all_data 
+sites <- unique(all_data$site)
+params <- unique(all_data$parameter)
+# Constructing a df to iterate over each site-parameter combination
+combinations <- crossing(sites, params)
+
+# Function to summarize site-parameter combinations
+summarize_site_param <- function(site_arg, parameter_arg, api_data) {
```
steeleb, line 314: praise: this is an excellent use of functions/prints/etc. 

suggestion: because this applies to all parameters across all sites, move this up to a section before we look at any individual site/parameters. Its current location doesn't make sense, IMO.

src/qaqc/01-organize_raw_data.Rmd
steeleb: thought: for the purposes of {targets}, I'd keep the flagging for sonde out of water separate from this function.

The workflow would be (IMO): 1) collate data, 2) recode where out of water, 3) apply the summarization in this function... and then we'll probably have some global flags that are applicable to parameters across all sites, and then, finally, site/parameter specific funcitons.

src/qaqc/01-organize_raw_data.Rmd
steeleb: Note, for efficiency and {targets}, Id' leave these two separate.

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -322,14 +398,24 @@ temperature_archery <- temperature_archery %>%
...
+temperature_archery <- temperature_archery %>%
+  mutate(flag = if_else(p1 >= temp_max | p1 <= temp_min,
+    # true
+    ifelse(is.na(flag), "Out of bounds",
+           paste(flag, "Out of bounds", sep = "; ")),
+    # false
+    flag),
+         p1 = if_else(p1 >= temp_max | p1 <= temp_min, NA_real_, p1))
+
+# Here is where I would need to add the paste so that it doesn't replace what is already in the flag column completely
```
steeleb, line 410: We could also have different flag columns. Sometimes collapsing into a single flag column is *not* the best way to move forward. 

For example: maybe there is a flag_sonde; flag_temp; flag_do; etc.

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -322,14 +398,24 @@ temperature_archery <- temperature_archery %>%
...
 ```
 
 Site visits: 
 
-```{r}
+```{r site visits}
 # flag the start time of a site visit
 temperature_archery <- temperature_archery %>%
-  mutate(flag = if_else((!is.na(visit_comments)), 
+  mutate(flag = if_else((!is.na(visit_comments)),
```
steeleb, line 418: Nitpick: still part of issue #19 and should be in an earlier section.

steeleb (approved):
Nice Juan!

I'm approving this, but adding a ton of comments. Given the current development of this pipeline, none of these comments are blocking, but in the next iteration, we should move the sonde-out-of-water section to somewhere around line 212. 

Let me know if any of my comments are unclear! Happy to clarify.

src/qaqc/01-organize_raw_data.Rmd
kathryn-willi: YES I agree ... actually thinking about targets needs it's better to keep these separate.

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -156,12 +156,12 @@ field_notes <- read_excel("data/sensor_field_notes.xlsx") %>%
          site = tolower(site),
          season = year(DT_round)) %>%
   arrange(DT_round) %>%
-  # binary version of deployed/pulled options. may need to play around with this som
+  # binary version of deployed/pulled options. may need to play around with this some
   # more
   mutate(where = case_when(!is.na(sensor_pulled) & !is.na(sensor_deployed) ~ 0,
```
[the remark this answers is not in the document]
kathryn-willi: Thoughts on this... 

1) If there is a field note listed, the sensor was almost **certainly** removed from the water for maintenance, and then redeployed at the end of the visit. (Most often for cleaning and calibrations.)
2) If there is an X in the sensor_pulled column, this means the sensor was completely removed from the site, likely due to it being broken. 
3) To identify the true start and end DT for a field season, we can identify the first instance in a year that the sonde has an X identified in sensor_deployed, and the last instance in a year that the sonde has an X identified in sensor_pulled.

src/qaqc/01-organize_raw_data.Rmd
kathryn-willi: Agreed! I keep forgetting about {targets}. For {targets} we will want to keep each step in the QAQC process split up into their own functions, even regardless of whether they can be performed across all site-parameter combos or one at a time. I also agree 100% with B's overarching workflow steps he's laid out here.

PR #28, opened 2023-08-25
Partial solution to issue #22

### LINKED ISSUE #22: Lines of code 278-283 are hard to understand
chore/nitpick: these lines of code (278-283) are really hard to follow and I'll be 100% honest that I have no idea what is happening here. Probably a me problem and likely could figure it out if I was running the code, but just reading/reviewing it would be helpful to have a comment before this so that the function is explained.

_Originally posted by @steeleb in https://github.com/rossyndicate/poudre_sonde_network/pull/18#discussion_r1294888110_

juandlt-csu:
This PR replaces the `setnafill()` method that was used to fill in NAs in the `where` column of `field_notes` with `fill()` from dplyr from issue 22. 

The exact location where `setnafill()` was used from the original issue 22 has not been replaced yet. Instead, this step has been added onto the `Load field notes` step, where the field notes data has been padded and rearranged so that this new `field_notes` df can be joined onto `all_data` earlier in the data pipeline. 

Note: both this and the previous method that was used are only accurate if the original field notes excel sheet is accurate. That should be checked to make sure that the `sonde_pulled` and `sonde_deployed` columns are accurate. For this reason I have decided to omit the option `.direction = "downup"` from `fill()`, for now.

This is progress on #22.

### DIFF
--- a/src/qaqc/01-organize_raw_data.Rmd
+++ b/src/qaqc/01-organize_raw_data.Rmd
@@ -155,13 +155,17 @@ field_notes <- read_excel("data/sensor_field_notes.xlsx") %>%
          DT_join = as.character(DT_round),
          site = tolower(site),
          season = year(DT_round)) %>%
-  arrange(DT_round) %>%
-  # binary version of deployed/pulled options. may need to play around with this some
-  # more
+  arrange(site, DT_round) %>% 
+  group_by(site) %>%
+  # this will pad DT_round grouped by site to determine when the sonde was deployed for that site
+  pad(by = "DT_round", interval = "15 min") %>%
+  # `where` determines if the sonde is deployed or not. 0 = sonde deployed, 1 = sonde is not deployed
   mutate(where = case_when(!is.na(sensor_pulled) & !is.na(sensor_deployed) ~ 0,
                            !is.na(sensor_pulled) & is.na(sensor_deployed) ~ 1,
                            is.na(sensor_pulled) & !is.na(sensor_deployed) ~ 0,
-                           is.na(sensor_pulled) & is.na(sensor_deployed) ~ 0)) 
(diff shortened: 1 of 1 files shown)


kathryn-willi:
I think the only direction for the fill() function should be "down", as we only want to fill in missing values in the top-> down direction.

kathryn-willi (approved):
With the new fill() upstream, you should now be able to remove all the strange nafill code! :)

steeleb:
Hi Juan!

I've edited your PR text to remove the closure of #22. That issue requires commenting, not code to close. What needs to be more clear is that this section of code *only* is to flag when the sonde was deployed or retrieved, resulting in that 0/1 for sonde being in use in the field. Because this type of flagging is only pertinent to rows of the field notes where there are x's in the deployed/pulled columns not all rows of the field notes, this is still confusing. 

Functionally, this works as intended - though maybe only starting in 2022 and/or at certain sites?

One nit pick: I think that field notes object should be renamed when you expand it to include all timestamps. When you do this, it is no longer just field notes. ;-)

I'm going to request that you just tag KW in future PRs unless you need my review for specific items/tasks. Great job and keep rolling with this!

steeleb (commented):
Hi Juan!

I've edited your PR text to remove the closure of https://github.com/rossyndicate/poudre_sonde_network/issues/22. That issue requires commenting, not code to close. What needs to be more clear is that this section of code only is to flag when the sonde was deployed or retrieved, resulting in that 0/1 for sonde being in use in the field. Because this type of flagging is only pertinent to rows of the field notes where there are x's in the deployed/pulled columns not all rows of the field notes, this is still confusing.

Functionally, this works as intended - though maybe only starting in 2022 and/or at certain sites?

One nit pick: I think that field notes object should be renamed when you expand it to include all timestamps. When you do this, it is no longer just field notes. ;-)

I'm going to request that you just tag KW in future PRs unless you need my review for specific items/tasks. Great job and keep rolling with this!

Issue #39, opened 2023-09-13
Create weekly plots per site/parameter combo


kathryn-willi:
Add the nearest upstream/downstream to the plot

PR #43, opened 2023-09-14
Adding weekly visualization function for flagged data

juandlt-csu:
This PR addresses issue #39 and introduces some simple changes:

- Created functional methods for visualizing flagged data.
    - Developed a separate function for daily visualization: `generate_daily_flag_plots()`
    - Developed a separate function for weekly visualization: `generate_weekly_flag_plots()`
        - This works, but I would classify it as buggy. Sometimes the plot will not filter properly (I think) and as a result not display all the days in the week in the plot. I explored why this happens (NA values, dates between months) but I couldn't figure it out. Something to look into for the future.
    - Developed function to stack the results from these functions: `weekly_daily_flag_plots()`
- Fixed a bug related to the full join of field notes.
    - Filtered field notes by site before joining them to the rest of the data. Weird things were happening without this step because of the full_join() step.
- Included a backup weekly visualization method in the scratch folder.

Apologies for the lengthy PR!

### DIFF
--- a/scratch/functions/generate_weekly_flag_plots.R
+++ b/scratch/functions/generate_weekly_flag_plots.R
@@ -1,0 +1,97 @@
+# Visualizing weekly data ----
+sites_order <- c("tamasag", # rist
+                 "legacy",
+                 "lincoln",
+                 "timberline",
+                 "prospect",
+                 "boxelder", # elc
+                 "archery",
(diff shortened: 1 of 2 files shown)


juandlt-csu:
There was a mistake in the code that I want to fix before I leave.

PR #45, opened 2023-09-18
Fix to site_vist/sonde_deployment flag

kathryn-willi:
Hey Juan,

I think I fixed the issue with the sonde deployment identifier, and the site visit identifier. I also wasn't able to continue running the code starting at line 452 - perhaps a branching problem? Mid-workflow problem? Otherwise, lookin' great.

Please let me know if the code I added makes sense, and works on your end! I think it's doing what I think...

### DIFF
--- a/src/qaqc/01-organize_raw_data.Rmd
+++ b/src/qaqc/01-organize_raw_data.Rmd
@@ -177,7 +177,8 @@ deployment_record <- field_notes %>%
                            !is.na(sensor_pulled) & is.na(sensor_deployed) ~ 1,
                            is.na(sensor_pulled) & !is.na(sensor_deployed) ~ 0,
                            is.na(sensor_pulled) & is.na(sensor_deployed) ~ NA),
-         last_site_visit = DT_round) 
+         last_site_visit = DT_round) %>%
+  filter(!is.na(sonde_employed))
 ```
 
 Munge field notes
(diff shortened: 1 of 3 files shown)


src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -380,13 +386,13 @@ all_data_summary_stats_list <- map(all_data_summary_list, generate_summary_stati
...
-# flag addition function
+# flag addition function 
 # This function will be called inside of add_x_flags() functions (ex. add_field_flags())
 # This function will be used to simply add flags into the flag column
 add_flag <- function(df, condition_arg, description_arg) {
   df <- df %>% mutate(flag = case_when(
-    {{condition_arg}} ~ if_else(is.na(flag), description_arg, paste(flag, description_arg, sep = ";\n")),
-    TRUE ~ flag))
+    {{condition_arg}} ~ if_else(is.na(flag), paste(description_arg), paste(flag, description_arg, sep = ";\n")),
+    TRUE ~ flag)) # <- KW question: what is this part of your code doing?
```
juandlt-csu, line 395: case_when() evaluates for the first `TRUE` condition in your case when pipe. This just hard codes `TRUE` at the end so that if none of the previous conditions were met you just return what is already in the flag column.

juandlt-csu (approved):
Hey Katie, this PR looks good to me! I like your solutions. 

Also, I will make a quick PR that should make the code from 452 on work. I can see how that's very unclear. Workflow issue I think. I was using the archery temp object as my testing df but then I switched to just adding the flags to all dfs so I could see them easier. I will try to be quick about that so you can use that code properly! Sorry about that!

PR #46, opened 2023-09-19
Cleaning up some objects and clarifying what the code is doing. 

juandlt-csu:
This PR cleans up the code so that there is no issue with running the code in the order that it is in. I also added some notes and comments to clarify what the code is doing at each step. Finally I commented out the cal_table step since we are not using that right now. This does not close any issues.


kathryn-willi (changes requested):
Not sure what's going on here, but maybe not all commits are included? I still don't see where `temperature_archery_bound_flags` gets created.

kathryn-willi (approved):

Issue #41, opened 2023-09-13
Histograms of all parameter/site combos


kathryn-willi:

juandlt-csu:
I forget what the goal for this issue was... What do we want to use these histograms for?

kathryn-willi:
To see what the range of values for a given parameter are. For example, we can use a histogram to see how frequently turbidity was above 100 NTU at a given site. Can be used to determine what our thresholds should be in a visual way!

juandlt-csu:
### Tasks:

- add (n = number of measurements) to the title of the histograms.
- add data subfolder for the images that are produced.
- change the months for the seasonal histograms so that they align with hydrology related seasons.
    - snowmelt = 4,5,6
    - monsoon = 7,8
    - fall baseflow = 9,10,11
    - winter baseflow = 12,1,2,3

juandlt-csu:
The histograms have been made and are ripe for exploration!

PR #51, opened 2023-09-26
adding n to title and altering the season months to align with hydrology

juandlt-csu:
This PR is progress on #41.

Quick PR that adds n = number of observations to histogram title and alters the season months for the seasonal histogram function.

### DIFF
--- a/src/qaqc/01-organize_raw_data.Rmd
+++ b/src/qaqc/01-organize_raw_data.Rmd
@@ -706,7 +706,8 @@ Histograms of all parameter/site combos
 generate_general_histogram <- function(df, df_index) {
   
   site_param <- toupper(sub("-", " ", df_index, fixed = TRUE))
-  title <- paste("Histogram of ", site_param)
+  n <- as.character(sum(!is.na(df$mean)))
+  title <- paste0("Histogram of ", site_param, " (n = ", n, ")")
   
   # there should be checks for data here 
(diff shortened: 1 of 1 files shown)


src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -743,21 +744,23 @@ temperature_hist_data <- map(temperature_hist_plots, ~ ggplot_build(.x)$data[[1]
 
 generate_seasonal_histogram <- function(df) {
   
-  winter <- c(12,1,2)
-  spring <- c(3,4,5)
-  summer <- c(6,7,8)
-  fall <- c(9,10,11)
-  seasons <- data.frame(winter, spring, summer, fall)
+  winter_baseflow <- c(12,1,2,3)
+  snowmelt <- c(4,5,6,NA)
```
SamStruthers, line 748: What are these NAs for?

juandlt-csu: They make each vector the same length so that they play nice when they are combined into a df. There shouldn't be any NA's in the month column since its derived from the DT_round column so I don't think there should be any issues with this...

SamStruthers: gotcha, makes sense

SamStruthers (approved):
Thanks for making the changes. @kathryn-willi unless you see any issues, I think its good to merge

kathryn-willi (approved):

PR #54, opened 2023-09-28
Added more flags, using seasonal data for thresholds

kathryn-willi:
All code in the QAQC up until the visualization steps has been tested and works. Includes the additions we did together as well as a few reorganization pieces at the flagging step.


src/qaqc/01-organize_raw_data.Rmd
juandlt-csu, line 1: This looks good and is in line with what we talked about yesterday. I will do some exploration to make sure that this is behaving correctly.

juandlt-csu (approved):
I appreciate you cleaning this up before sending it out to me! Looks good to me.

PR #56, opened 2023-10-04
Edits, function additions, and flag exploration

juandlt-csu:
This PR does not close any issues. Here is the state that I am leaving this project after some smaller edits and a lot of data exploration. I would appreciate some help with the large anomaly flag function.

- Added  sv window to site visit flag in order to flag the following 45 minutes after a site visit
- Added large anomaly flag function (help here would be appreciated)
    - The idea here is that this function would flag large swaths of bad data. This is done in case there is a lot of bad data followed by reasonable data in a window. We want to flag the reasonable data within that window because it has become suspect (hard to trust it after so much weirdness in the data). This is done by:
        - Creating the `flag_binary` column to track if there are flags in the flag column.
        - Then creating the `roll_bin` column which sums up the `flag_binary` for a 24hr rolling window, with that point as the center.
        - Finally this function adds the “24hr anomaly flag” to the flag column if a quarter of the data in a 24hr rolling period has flags and returns the df.
    - Issues I am having:
        - Extremely inclusive (97 data point (24 hour) rolling window).
        - I thought that by setting the condition `flag == "24hr anomaly flag”` to 0 it would make it less inclusive, but it is still too inclusive to be effective, I think.
        - Another idea would be to reduce the window size, but I haven’t experimented with that yet.
- Altered the generating daily plots function 
    - Added if else statements to bypass some obvious errors when iterating over all the data.
    - Added y_min and y_max hlines to the plots that are generated to easily discern when points are out of seasonal ranges. 
    - Sorted the output of the function.
- Added new histograms to data folder
    - Same as before, just updated with the new seasons that we are using and adding vlines to see how much of the data is being excluded with the the quantiles we are using (10&99).


src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -302,12 +309,21 @@ add_range_flags <- function(df) {
...
     # adding lab bound flags
-    add_flag(parameter == parameter_name & (mean < m_mean10 | mean > m_mean90),
+    add_flag(parameter == parameter_name & (mean < m_mean10 | mean > m_mean99),
              paste0("outside of ", parameter_name," seasonal range"))
   
   return(df)
   
 }
+
+# exploring data ----
```
steeleb, line 319: Note: `all_data_flagged` isn't created until line 439!

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -250,9 +253,11 @@ Add flagging functions for each df in all_data_summary_list
 # This function will be called inside of add_x_flags() functions (ex. add_field_flags())
 # This function will be used to simply add flags into the flag column
 add_flag <- function(df, condition_arg, description_arg) {
+  # adding the flag
   df <- df %>% mutate(flag = case_when(
     {{condition_arg}} ~ if_else(is.na(flag), paste(description_arg), paste(flag, description_arg, sep = ";\n")),
-    TRUE ~ flag)) # <- KW question: what is this part of your code doing? 
+    TRUE ~ flag)) 
+  
   return(df)
```
steeleb, line 261: as a general note, per {tidyverse} style guide, you only need to use `return()` if it is an 'early return' in a function. 
```suggestion
 df
```

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -383,16 +394,22 @@ add_slope_flags <- function(df) {
 # }
 ```
 
+add large anomaly flag function
+```{r}
+add_large_anomaly_flag <- function(df) {
+  df <- df %>% 
+    mutate(flag_binary = ifelse((is.na(flag) | flag == "24hr anomaly flag"), 0, 1),
+           roll_bin = data.table::frollsum(flag_binary, n = 97, align = 'center', na.rm = F, fill = NA_real_)) %>%
```
steeleb, line 402:
```suggestion
           # create a rolling count of flags over the 12-h prior and 12-h after the current row of data
           roll_bin = frollsum(flag_binary, n = 97, align = 'center', na.rm = F, fill = NA_real_)) %>% 
```
no need to use package name here, no conflicting function names - in an effort to make more literate code, I'd also suggest renaming this to something that helps an outsider know what this column is, for instance 'rolling_flag_count' or the like.

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -383,16 +394,22 @@ add_slope_flags <- function(df) {
 # }
 ```
 
+add large anomaly flag function
+```{r}
+add_large_anomaly_flag <- function(df) {
+  df <- df %>% 
+    mutate(flag_binary = ifelse((is.na(flag) | flag == "24hr anomaly flag"), 0, 1),
```
steeleb, line 401: Add comments here to clarify what each step is doing, please.
Suggestion: use dplyr::if_else() instead of base-r ifelse().

```suggestion
    # add an indication where there is a flag, except when the flag is the 24h anomaly.
    mutate(flag_binary = if_else((is.na(flag) | flag == "24hr anomaly flag"), 0, 1),
```

src/qaqc/01-organize_raw_data.Rmd
```diff
@@ -383,16 +394,22 @@ add_slope_flags <- function(df) {
...
+           roll_bin = data.table::frollsum(flag_binary, n = 97, align = 'center', na.rm = F, fill = NA_real_)) %>% 
+    add_flag((roll_bin >= (97*0.25)), "24hr anomaly flag")
```
steeleb, line 403:
```suggestion
    # add flag for untrusworthy data if the total number of rows of data that have flags is greater than 25% of all rows
    add_flag((roll_bin >= (97*0.25)), "24hr anomaly flag")
```

steeleb (commented):
Looking good!

I made a few suggestions to help nudge you towards {tidyverse} style guide compliance.

Otherwise your large anomaly flag looks solid, I think because your other flags are... flagging a lot of things... this is going to be a greedy flag. Keep working on getting those other flags to do what you want them to, and this one will be a better indicator of generally-problematic data. 

A note: be careful about using environment variables, even in scratch code, that don't yet exist in the R environ, it makes it pretty difficult for others to hop in to help troubleshoot, as we have to trouble shoot that first. 😉

Finally, it might be time to move the compilation of the data files to another script and read in the data at the top of this one. This is purely my personal preference because I hate waiting for code to compile. 

Keep on keeping on!

scratch/misc/percent_flagged.R
```diff
@@ -0,0 +1,82 @@
...
+                     "anomaly window")
+
+  # summarize total data points
+  total_observations <- df %>%
+    summarise(n_total = n_distinct(DT_round)) %>%
+    pull(n_total)
+
+  # summarize total days
+  total_observations_dates <- df %>%
+    group_by(date = format(DT_round, "%m-%d-%Y")) %>%
```
steeleb, line 26: Suggestion: probably best to format dates as YYYY-MM-DD. KW might have thoughts on this, but generally it's a format that is autodetected. 

You can also use {lubridate} here:
`group_by(date = ymd(DT_round)) %>%`

scratch/misc/alter_flag.R
```diff
@@ -0,0 +1,8 @@
+# Flag alter function
+# This function can be used to alter flags that have been established
+alter_flag <- function(df, condition_arg, old_description_arg, new_description_arg) {
+  df <- df %>% mutate(flag = case_when(
+    {{condition_arg}} & str_detect(flag, old_description_arg) ~ str_replace(flag, old_description_arg, new_description_arg),
+    TRUE ~ flag))
+  return(df)
```
steeleb, line 7: {tv} sytle guide: `return()` unnecessary

scratch/misc/percent_flagged.R
```diff
@@ -0,0 +1,82 @@
...
+    summarize(n_total = nrow(date)) %>%
+    nrow()
+
+  row_list <- list()
+  for (i in list_of_flags) {
```
steeleb, line 31: I challenge you to functionalize this! :)

steeleb: you could either do that as a separate function that you call here 
OR! 
you can map and define the function:
`map(list_of_flags, function(x){` 
and then mostly replace `i` with `x` (and a few other tweaks, but I think you get where I'm going here!)

scratch/misc/percent_flagged.R
```diff
@@ -0,0 +1,82 @@
...
+    percent_flagged_dates <- flagged_observations_dates/total_observations_dates
+
+    # creating a vector with the information
+    calculated_values <- tibble(
+      # metadata
+      site = site,
+      parameter = parameter,
+      flag = i,
+      # data points
+      data_points_flagged_percentage = format(percent_flagged * 100, nsmall=2),
```
steeleb, line 56: you could also use `round()` here.

scratch/misc/percent_flagged.R
```diff
@@ -0,0 +1,82 @@
...
+
+    row_list[[i]] <- calculated_values
+  }
+
+  #calculated_df <- bind_cols(row_list)
+  return(bind_rows(row_list))
+
+}
+
+flag_report <- imap(all_data_flagged, ~generate_flag_report(.x,.y)) %>%
```
steeleb, line 73: you're a legend in mapping. imap makes my brain hurt. well done!

steeleb (commented):
Looking good! 

For each of these functions that I assume is on the path to {targets}-ifying, make sure you add formal documentation to each of them. We haven't added it to our best practices yet, but you can see it at work in the aquasatv2 and lakeSR repos in the `src` folders that contain the R functions called in the {targets} workflow.

PR #57, opened 2023-10-18
added ability to pass or fail an entire day's worth of data in remove…

juandlt-csu:
This PR just adds the ability to pass or fail an entire day's verification column within remove_flag_method.R. Also simplifies some of the code that was in the function.

### DIFF
--- a/src/qaqc/remove_flag_method.R
+++ b/src/qaqc/remove_flag_method.R
@@ -2,11 +2,30 @@
 ## This uses the data that is generated from the function generate_daily_flag_plots()
 ## to ask the user whether or not a flagged data point should pass or fail.
 
+# get day decision function ----
+# returns "pass", "fail", or "inspect depending on input from the user
+get_day_decision <- function(prompt_text) {
+  while (TRUE) {
+    user_input <- readline(prompt = paste(prompt_text))
+    user_input <- tolower(user_input)
+
+    if (user_input %in% c("pass", "p")) {
+      return("pass")
+    } else if (user_input %in% c("fail", "f")) {
+      return("fail")
+    } else if (user_input %in% c("inspect", "i")){
+      return("inspect")
+    } else {
+      cat("Invalid input. Please enter 'yes', 'no', or 'inspect'.\n")
+    }
+  }
+}
+
 # get flag decision function ----
 # returns a binary TRUE or FALSE depending on the input from the user
 get_flag_decision <- function(prompt_text) {
   while (TRUE) {
-    user_input <- readline(prompt = paste(prompt_text, " (pass/fail): "))
+    user_input <- readline(prompt = paste(prompt_text))
     user_input <- tolower(user_input)
 
     if (user_input %in% c("pass", "p")) {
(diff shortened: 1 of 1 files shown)


src/qaqc/remove_flag_method.R
```diff
@@ -48,81 +46,77 @@ alter_verification_column <- function(df, dt, choice_arg = choice) {
...
+                       color = "data point of interest",
+                       # size = 2, # fix this later
+                       alpha = 0.05))
+
+      # Print the plot so the user can see the data
+      print(ggarrange(week_plot, day_plot, nrow = 2, ncol = 1))
+
+      # Generate the prompt for the user to answer
+      flag_dt <- as.character(flag_data$DT_round[i])
+      flag_choice_prompt <- paste(flag_dt, " (pass/fail): ")
```
SamStruthers, line 104: This seems like a lot of prompting... 15 min data for 1 day is like 100 data points. Can we make a prompt where you can select a few hours instead?
I realize that would be a bit complicated but reviewing 100 data points per sensor per site would hurt my brain...

src/qaqc/remove_flag_method.R
```diff
@@ -133,15 +127,14 @@ verify_flag_data <- function(weekly_plot_object, daily_plot_object) {
 # Generating the list of plots (this example only has one plot that has a repeated value flag)
 # we are doing it this way so that we can get the data from the daily ggplot
 test_weekly <- generate_weekly_flag_plots("archery", "Temperature", "slope violation")
```
SamStruthers, line 129: My selecting a specific flag, could we be verifying the same data multiple times and redoing the work again? I wonder if the function above can also remove any plots which already have verification == TRUE

src/qaqc/remove_flag_method.R
```diff
@@ -133,15 +127,14 @@ verify_flag_data <- function(weekly_plot_object, daily_plot_object) {
...
 # maybe we will use map2() so that it can take the input from both and then combine them?
-verify_flag_data(test_weekly[[2]], test_daily[[2]])
+verify_flag_data(test_weekly[[4]], test_daily[[4]]) %>% View()
 
 # Using the function that was made to map over the dfs in the list
 # make sure to put in the weekly information first
-test_map <- map2(head(test_weekly), head(test_daily), verify_flag_data)
-
-# For now only the daily plots are generated, but this can change in the future.
+# test_map <- map2(head(test_weekly), head(test_daily), verify_flag_data)
```
SamStruthers, line 140: I like this idea!

SamStruthers (approved):
I think the general concept is great and code seems to make sense to me. 
We should just think about the usability of this function and optimize to make sure we are not creating too much work for ourselves

src/qaqc/remove_flag_method.R
juandlt-csu: This is true! 

To circumvent re-reviewing data the weekly and daily plot generation functions have the ability to print all the plots with flags if the flag_arg is NULL. This makes it so that if any flag is found in a date, it will just print that plot without the need to specify a flag. Since I am testing I wanted to print out some of the flags and not all of the flags for speed.

I do think that we will still need to make a simple system that tracks which data has been reviewed and which data has not been reviewed, though. A column that has verification == TRUE, or something along those lines is a good idea.

src/qaqc/remove_flag_method.R
```diff
@@ -48,81 +46,77 @@ alter_verification_column <- function(df, dt, choice_arg = choice) {
...
+  day_choice_prompt <- paste("Would you like to (pass/fail/inspect) all data points for: ", df_name, "? (p/f/i): ")
+  day_choice <- get_day_decision(day_choice_prompt)
+
+  # alter the df based on the response from the user
+  if (day_choice == "pass") { # if choice is true
+    # add pass to verification column for all rows
+    altered_df <- df_data %>%
+      mutate(verification = "pass")
+    cat("All points for ", df_name, "have PASSED.\n")
+    return(altered_df)
```
kathryn-willi, line 67: I don't think you need the `return()`'s in if else statements.

kathryn-willi (approved):

juandlt-csu:
Something to fix: As it it datapoints that are NA automatically fail. If a user approves an entire day, then NA data points that had previously failed will be converted to passes, which is wrong. This is incorrect and should be fixed.

kathryn-willi:
Good catch, Juan. I also think that we will need to split up flags related to the sensor giving out bad data, and flags related to data not being collected. (Will discuss this/formalize in GitHub Issues tomorrow during our meeting)

PR #58, opened 2023-10-19
Automating field notes using mWater

SamStruthers:
mWater is a survey platform used by CU folks to record field note visits. Data is stored in a datagrid associated with a survey on the mWater portal. These data can be downloaded using an API url (mWater api yml)
This PR has three main files which all live in the folder `mWater_collate`:
- `mWater_pull_collate.rmd`: main file which pulls data from mWater API and breaks it into three main sections: all_notes, sensor_notes, sampling_notes. 

Dataframes: All notes is everything submitted, sensor_notes is only visits where sensor was interacted with, sampling notes is when water sampling occured.  

Actions: save sensor notes to csv in data folder this is how we can maintain a backup of the data from mWater. This code also uses the photo_download and sampling spreadsheet creator functions (described below)
- `photo_downloader.R`: grabs all the download urls from the all notes sheet and will download+rename all the photos uploaded to mWater. It should not download any photos if they have already been downloaded
- `sampling_spreadsheet_creator.R`: formats a days worth of water sampling data to match RMRS formatting. This is mainly a tool that I will use after a field day to speed up sample input and make sure I don't create any typos in the RMRS masterfile. 

Review Requested:
- I will send you the API yml. Please test to make sure the whole pipeline works by running all the chunks (keeping in mind that photo downloader wont do anything since I already downloaded the photos)
- The code is not very concise, especially photo downloader so any suggestions on that would be helpful
- Check to make sure I mimic'ed the existing structure of our field notes in QAQC pipeline (start_dt, round_dt, sensor employed, etc)
- general notes on conciseness of rmd file and any sections that seem confusing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ data
 .DS_Store
 images/
 .DS_Store data
+src/mWater_collate/mWater_API.yml
--- a/src/mWater_collate/download_pictures.R
+++ b/src/mWater_collate/download_pictures.R
@@ -1,10 +1,10 @@
-library(tidyverse)
+
 
 download_pictures <- function(){
   # Find all the downloaded pictures
   all_file_names <- list.files(path = "data/field_pics/", recursive = TRUE)
   #grab notes
-  sampling_photos <- all_notes%>%
+  sampling_photos <- all_notes_cleaned%>%
     #grab needed columns
     select(site, start_dt,photos_downloaded,upstream_pic,downstream_pic,clarity,filter_pic,other_pic,other_pic_descriptor)%>%
     mutate(
@@ -54,7 +54,7 @@ path <- "data/field_pics/"
 
 
   # loop thru dataset and download the photo ONLY if it is not yet downloaded and not NA
-  for (i in 1:length(sampling_photos$site)) {
+  for (i in 1:nrow(sampling_photos)) {
     if (!is.na(sampling_photos$upstream_downloaded[i]) && !sampling_photos$upstream_downloaded[i]) {
       download.file(sampling_photos$upstream_pic[i], destfile = paste0(path,sampling_photos$upstream_filename[i]))
     }
@@ -71,7 +71,7 @@ path <- "data/field_pics/"
   }
 
   #grab notes for sites with other pictures
-  other_photos <- all_notes%>%
+  other_photos <- all_notes_cleaned%>%
     #grab needed columns
     select(site, start_dt,other_pic,other_pic_descriptor)%>%
     #get rid of instances with no other pics
@@ -98,10 +98,10 @@ path <- "data/field_pics/"
            ))
 
   # loop thru dataset and download the photo ONLY if it is not yet downloaded and not NA
-  for (i in 1:length(other_photos$site)) {
+  for (i in 1:nrow(other_photos)) {
     if (!is.na(other_photos$other_downloaded[i]) && !other_photos$other_downloaded[i]) {
       download.file(other_photos$other_pic[i], destfile = paste0(path,other_photos$other_filename[i]))
     }}
 
-
+cat("All Available Pictures Downloaded")
 }
--- a/src/mWater_collate/download_pictures.R
+++ b/src/mWater_collate/download_pictures.R
@@ -1,0 +1,107 @@
+library(tidyverse)
+
+download_pictures <- function(){
+  # Find all the downloaded pictures
+  all_file_names <- list.files(path = "data/field_pics/", recursive = TRUE)
+  #grab notes
+  sampling_photos <- all_notes%>%
+    #grab needed columns
+    select(site, start_dt,photos_downloaded,upstream_pic,downstream_pic,clarity,filter_pic,other_pic,other_pic_descriptor)%>%
+    mutate(
+      #Date format for pictures
+      yyyymmdd = format(start_dt, "%Y%m%d"),
+      #create filenames ONLY if there is a URL associated with the site visit
+      upstream_filename = case_when(
+        !is.na(upstream_pic) ~ paste0(site, "_", yyyymmdd, "_upstream.jpg"),
+        TRUE ~ NA_character_
+      ),
+      downstream_filename = case_when(
+        !is.na(downstream_pic) ~ paste0(site, "_", yyyymmdd, "_downstream.jpg"),
+        TRUE ~ NA_character_
+      ),
+      clarity_filename = case_when(
+        !is.na(clarity) ~ paste0(site, "_", yyyymmdd, "_clarity.jpg"),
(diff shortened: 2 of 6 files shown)


.gitignore
juandlt-csu, line 1: I was able to use the .yml file for its intended purpose

src/mWater_collate/download_pictures.R
```diff
@@ -0,0 +1,107 @@
...
+        downstream_filename %in% all_file_names ~ TRUE,
+        TRUE ~ FALSE
+      )
+    )
+# basic path to field pics
+path <- "data/field_pics/"
+
+
+  # loop thru dataset and download the photo ONLY if it is not yet downloaded and not NA
+  for (i in 1:length(sampling_photos$site)) {
```
juandlt-csu, line 57: suggestion: If this is iterating over the number of rows in the df, I think it would be more clear to use `for (i in 1:nrow(sampling_photos))`.

src/mWater_collate/mWater_pull_collate.Rmd
```diff
@@ -0,0 +1,166 @@
...
+date: "`r Sys.Date()`"
+output: html_document
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE)
+library(tidyverse)
+
+#grab site metadata
+sampling_sites <- read_csv("data/water_sampling_sites.csv")
```
juandlt-csu, line 13: I had to change the file location here, but I don't know why... Generally I was struggling with file locations. I don't know if this is a problem with this PR or with my local instance. Will have to check this again once it is merged.

src/mWater_collate/mWater_pull_collate.Rmd
juandlt-csu, line 1: I think that this is generally working correctly, but I would appreciate some documentation that explains overall what should be happening.

src/mWater_collate/mWater_pull_collate.Rmd
```diff
@@ -0,0 +1,166 @@
...
+
+```
+
+
+# Tidying
+
+## ALL NOTES
+```{r}
+#read in file
+all_notes <- read_csv(file = "data/mWater_raw_sensor_notes.csv")%>%
```
juandlt-csu, line 38: Which one of these dfs are we hoping to append to the field notes sheet that we already have?

src/mWater_collate/mWater_pull_collate.Rmd
```diff
@@ -0,0 +1,166 @@
...
+sensor_notes <- all_notes%>%
+  filter(grepl("Sensor",visit_type))%>%
+  select(-c(chla_volume_ml, vol_filtered_blank_dup, sample_collected,do_mgl,
+            cond_ms_cm, temp_c, upstream_pic, downstream_pic, clarity, other_pic, other_pic_descriptor))%>%
+  mutate(sonde_employed = case_when(is.na(sensor_change)  ~ 1,
+                                    sensor_change == "Pulled" ~ 0,
+                                    sensor_change %in% c("Swapped", "Deployed") ~ 1), 
+         # Create basis for calibration report name
+         # this will be used to check for calibration report in data files and then b
+         cal_report_name = case_when(cal_report_collected == TRUE ~ paste0(site, "_", format(start_dt, "%Y%m%d")),
```
juandlt-csu, line 79: question: what is the log stuff doing? Does mWater have a way to download and interact with the calibration report stuff?

src/mWater_collate/sampling_spreadsheet_creator.R
```diff
@@ -0,0 +1,21 @@
+sampling_spreadsheet_creator <- function(date_oi = "2023-10-16" ){
+
+  #pull in site meta data
```
juandlt-csu, line 3: quibble: I like to add lines between steps/chunks of code to improve readability.

juandlt-csu (approved):
This seems to be working correctly to me. Although, I thought that the end goal would be to generate a df from the mWater field notes that are pulled such that it can be appended to the field notes that we already have. It wasn't clear to me which generated df would be the one that would be appended to our historical field notes.

src/mWater_collate/download_pictures.R
SamStruthers: Will do, thanks!

SamStruthers:
Thanks for the notes. I'll make some changes to the scripts to make it clear that sensor_notes is what we will be pulling into the QAQC script.

src/mWater_collate/mWater_pull_collate.Rmd
SamStruthers: sensor_notes
apologies that was not clear

src/mWater_collate/mWater_pull_collate.Rmd
SamStruthers: Not yet! I want to build out a way to make sure that techs/us are uploading all our files in a timely manner. This will essentially be the file that R will look for in the data folder to see if the file has been uploaded or not

SamStruthers:
Big Improvement (IMO): 
- Improved API pull (no raw mWater file anymore)
- all_notes_cleaned is the df that is used downstream and only has a few minor corrections from API pull
- readability is improved and sensor notes is the only other dataframe created
- Finished log/cal report check function: This checks the data folder to see if users have uploaded the logs/cal reports that they claim to in the mWater survey
- Moved all water sampling dataframes into the sampling spreadsheet function as they are not needed in the main rmd. Users can now save all the water sampling info as csv or just a day in RMRS formatting
- Fixed the nrow comment in download pictures.R

src/mWater_collate/mWater_pull_collate.Rmd
kathryn-willi: I also had issues with the file locations... not sure why...

kathryn-willi (approved):
This is awesome, Sam! And going to make field days much easier. @juandlt-csu, we will need to come up with a way of appending our old field notes to this newer data at the top of our field notes script in the QAQC workflow. We will also want to confirm all the original column names line up. On that topic, I will go ahead and add in a `visit_type` column to our old data and fill that column in so that we can use that field to filter/remove visits where the sensor wasn't actually handled.

SamStruthers:
I added a new file `scratch/data_tidying/timelapse_folder_rename.R`
This is mostly an tool that I will use when I get SD cards from our timelapse cameras. 
It essentially prompts the user for the site name after showing a picture from the folder in the `timelapse_raw` folder. Based on the response and the first and last photo in the folder, it renames the folder to match our syntax and then moves said folder to the `timelapse_photos` folder. 

No review required since I will be the main user of this and it does not affect our larger workflow unless anyone feels strongly about this....

PR #66, opened 2023-11-03
Minor Tweak to mWater pull Collate

SamStruthers:
Fixed the visit type other to be joined into visit type. 
- Removes other(please specify) with the value of visit type other
- date was being saved as UTC so samples or site visits after 6pm were being saved as the incorrect date which lead to issues in sampling spreadsheet creator

No in depth review required but run of code would be great to make sure it is working on other machines since we had issues with that.


src/mWater_collate/sampling_spreadsheet_creator.R
```diff
@@ -42,7 +42,7 @@ sampling_spreadsheet_creator <- function(date_oi = "2023-10-16", all_dates = FAL
    write_csv(x = sampling_notes, file = paste0("data/sampling_notes/all_samples_as_of_",as.character(Sys.Date()),".csv" ))
  }else{
    #grab desired date
-   date_oi_clean <- as.Date(date_oi)
+   date_oi_clean <- as.Date(date_oi, tz = "America/Denver")
```
juandlt-csu, line 45: Comment: I still don't get time zones very well...

src/mWater_collate/mWater_pull_collate.Rmd
```diff
@@ -52,7 +52,7 @@ all_notes_cleaned <- all_notes%>%
     #start and end dt comes in as UTC -> to MST
         start_dt = with_tz(ymd_hms(start_dt), tz = "America/Denver"),
         end_dt = with_tz(ymd_hms(end_dt), tz = "America/Denver"),
-         date = as.Date(start_dt),
+         date = as.Date(start_dt, tz = "America/Denver"),
```
kathryn-willi, line 55: is this going to always be in MST? Important to put in MST due to how API data comes in....

kathryn-willi (approved):
Looks good. Just ensure that DTs are always MST in the final product :)

SamStruthers:
Changed to MST :)

PR #78, opened 2023-11-14
New suspect data flag, new network check flag

kathryn-willi:
Lots of commits here... most are of little consequences. However! I did do a few really big things -

1) I added a new function in 'site_comp_test.R'. This is my fist pass at using network relationships to figure out if flags are occurring up/downstream of a given site, then using that as a means to confirm if a flag is warranted. This is used at the very end of the flagging workflow, and produces a new column called "cleaner_flag" (... feel free to change the name)

2) The slope flag is now based on the 1st and 99th percentile slope value, meaning the direction of slope (positive/negative) matters when we flag slope.

3) I added my own version of the anomaly flag, and renamed that flag to "suspect data". I tried to find all instances of the old flag's name across the entire project repo, but be careful! I may have missed some...


src/qaqc/download_and_flag_fxns/add_seasonal_flag.R
```diff
@@ -14,19 +14,15 @@ add_seasonal_flag <- function(df) {
...
-           parameter = parameter.x) %>%
-    # ... flag obs that are outside the seasonal 1-99 percentile range:
+    left_join(lookup, by = c("site","parameter","season")) %>%
+    # ... flag obs that are outside the seasonal 1st - 99th percentile range:
     add_flag((mean < t_mean01 | mean > t_mean99), "outside of seasonal range") %>%
-    # flag obs whose slope is greater than the 99th percentile range
-    add_flag((slope_ahead >= t_slope_behind_99 | slope_behind >= t_slope_behind_99), "slope violation") %>%
+    # flag obs whose slope is outside the 1st - 99th percentile range
+    add_flag((slope_ahead >= t_slope_behind_99 | slope_behind >= t_slope_behind_99 |
+              slope_ahead <= t_slope_behind_01 | slope_behind <= t_slope_behind_01), "slope violation") %>%
```
juandlt-csu, line 22: question: I thought that slope_behind_01 would be the tamest/smallest slopes? Were you able to generate the flag report after you did this to see if the percentage of slope violations increased?

src/qaqc/download_and_flag_fxns/generate_summary_statistics.R
```diff
@@ -32,7 +32,7 @@ generate_summary_statistics <- function(site_param_df) {
       rollsd = roll_sd(mean, n = 7, align = 'center', na.rm = F, fill = NA_real_),
       # Determine the slope of a point in relation to the point ahead and behind.
       slope_ahead = abs(front1 - mean)/15,
-      slope_behind = abs(mean - back1)/15,
+      slope_behind = (mean - back1)/15,
```
juandlt-csu, line 35: question: is this supposed to not be an absolute value anymore?

src/qaqc/download_and_flag_fxns/make_threshold_table.R
```diff
@@ -1,25 +1,26 @@
...
-  # by = "season")
+  good_data_stats <- df %>%
+    # REMOVE DATA WE KNOW TO BE ERRONEOUS:
+    add_malfunction_flag() %>%
+    filter(is.na(flag)) %>%
+    group_by(season) %>%
+    mutate(f01 = quantile(mean, 0.01, na.rm = TRUE),
+           f99 = quantile(mean, 0.99, na.rm = TRUE),
+           f_slope_behind_01 = quantile(slope_behind, 0.01, na.rm = TRUE),
+           f_slope_behind_99 = quantile(slope_behind, 0.99, na.rm = TRUE)) %>%
```
juandlt-csu, line 11: question: how do these quantiles work now that slope behind is not an absolute value?

I don't know if I understand this correctly, but if we wanted to differentiate between slopes going and slopes going down I would think that we want to separate the positive values and the negative values into their own columns and then get the 99th quantiles from those categories? This stems from my understanding that the 1st percentile of slopes would be the smallest slopes, but a negative slope wouldn't necessarily mean a small slope in this instance, since we are no longer taking absolute values.

src/qaqc/download_and_flag_fxns/add_seasonal_flag.R
kathryn-willi: Because we are no longer getting the abs(slope) the 1st percentile slope will actually be a negative number now, and therefore can be considered an extreme case of a negative slope...I think? Could you check that out for me?

src/qaqc/download_and_flag_fxns/generate_summary_statistics.R
kathryn-willi: Correct, I no longer want this to be the absolute slope value - see my comment above.

juandlt-csu (approved):
Hey Katie, this looks good to me. I think this looks good to push to main, I have mostly minor comments, except for that t_slope_behind_01 stuff (again I don't really know if I understood that correctly). PS the last of the things you pointed out (renaming the flag and finding all of the instances) is why I think we should focus on targetsifying now, I have been struggling a lot with this lately! Great PR :)

src/qaqc/download_and_flag_fxns/site_comp_test.R
```diff
@@ -0,0 +1,83 @@
...
+                   "river bluffs")
+
+  width_fun = ifelse(site_name == "tamasag", 20,
+              ifelse(site_name == "legacy",16,
+              ifelse(site_name == "lincoln", 16,
+              ifelse(site_name =="timberline", 16,
+              ifelse(site_name ==  "prospect", 12,
+              ifelse(site_name == "boxelder", 16,
+              ifelse(site_name ==   "archery", 12,
+              ifelse(site_name == "river bluffs", 12, NA))))))))
```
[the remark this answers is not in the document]
kathryn-willi: yes!

juandlt-csu:
I pushed this PR into my (Juan) main fork. It has a merge conflict with the main branch that I will resolve on my end before pushing it back into the main branch.

PR #85, opened 2023-11-28
Targets

juandlt-csu:
This PR updates the `targets` branch from our main repo with the `targets` branch that I have been working on. A lot of our function files have been changed so that they work within the targets systems, and I think that this means that they would not be functional in the .Rmd files, though I am not certain. I think that we are going to have to shift the structure of this project again so that it makes sense in the context of the FC system.

This works on my machine, if you want to run the pipeline I would recommend either deleting the files that are in "data/api/test_data/" and then running `tar_make(incoming_data_csvs)` (when I did this it took an hour to pull in all of the new data), OR (this one is better, I think) just commenting out the incoming_data_csvs tar_target().

Still a good amount of work to be done, but there is progress towards automation.


src/api_pull/api_puller.R
```diff
@@ -0,0 +1,74 @@
...
+  # Here, a way to find the most recent download of the data. Use this as the start date to
+  # reduce overlapping data
+
+  # tz weirdness
+  utc_start_date <- format(as.POSIXct(format(start_dt, tz = "UTC"), tz = "UTC"), format = "%Y-%m-%d %H:%M:%S")
+
+  utc_end_date <- format(as.POSIXct(format(end_dt, tz = "UTC"), tz = "UTC"), format = "%Y-%m-%d %H:%M:%S")
+
+  timezone = "UTC"
+
```
kathryn-willi, line 33: might be nice to add something like `print(paste0("No data at ", site, " during this time frame"))` for testing/checking purposes.

_targets.R
```diff
@@ -0,0 +1,184 @@
...
+          add_na_flag() %>%
+          add_repeat_flag() %>% 
+          kw_add_large_anomaly_flag() %>%
+          mutate(mean_public = ifelse(is.na(flag), mean, NA)) %>%
+          mutate(historical_data = TRUE)
+      })
+      # network check
+      final_flag <- map(all_data_flagged, site_comp_test)
+
+      all_data_flagged <- final_flag
```
kathryn-willi, line 180: will want to save this final product as the updated all_data_flagged object

kathryn-willi (approved):
Since the purpose of this PR is just to test that the targets pipeline doesn't throw errors.... hooray! It didn't for me. :) Next steps are definitely to ensure that the meat of the functions are doing what we want them to do. Namely, pulling in the old flagged data, appending it to the new data, and creating a new, updated flagged dataset.

src/qaqc/download_and_flag_fxns/generate_summary_statistics.R
juandlt-csu, line 1: The changes made in this file are because we are appending data that has been flagged to data that has not been flagged before the generate summary statistics step. The changes here make the function check if the columns we are updating are NA before we calculate their statistics. This ensures that data coming in will be updated only if it is empty, while also making sure that we are not losing data that has already been calculate at the head and tail of the df.

src/qaqc/01_download_and_flag.Rmd
juandlt-csu, line 1: These changes were mainly so that I could reload the flagged_data into the targets with the columns that I was working with. That being said I think that we should consider converting this RMD to our initial data upload for the FC system, while the targets pipeline will be what is run live.

PR #89, opened 2023-12-07
Targets + mWater notes integration

SamStruthers:
Purpose: mWater notes were not yet incorporated to targets workflow. 

Work: 
- Updated scripts in mWater_collate folder in Targets branch
- Functionalized `grab_sensor_notes.R` for targets workflow
- Added in `_targets.R`  xlsx notes are now `old_field_notes`,  pulled in `mWater_field_notes`, and binded them to create `field_notes`
- Ran the targets workflow again, (my run is test_all_data.RDS, sorry if I was supposed to change that before running...) nothing threw errors, but there seemed to be missing data (11/2 - 11/30) not sure if that is by design or something with API pull or new field notes?

Review:
- Check to make sure I didn't break it/ confirm missing data (11/2 - 11/30) was not due to new field notes integration
- Is my targets-ifying of `grab sensor notes` done in the best way?
- All other code was reviewed in PR to main a few days ago, just translating it to Targets


src/mWater_collate/grab_sensor_notes.R
```diff
@@ -0,0 +1,47 @@
+
+grab_mWater_sensor_notes <- function(){
+
+  #source clean mwater script for all notes cleaned
+
+  source("src/mWater_collate/clean_mwater_notes.R")
```
juandlt-csu, line 6: to do: If you source this at the beginning of the targets script I don't think you should have to source it here. If you can you should check if that is the case.

juandlt-csu (approved):
- I'll check the missing data issue once this is merged. 
- grab sensor notes will probably be changed in the near future now that the sensors are out. It will change because:
    - I think that we are going to have to make a historical set of flagged data for the initial data upload to the FC system (this will probably be done in the Rmd script that we had been using). 
    - I noticed that this function comments out a section that updates a csv file with the information that it pulled from mWater. We will probably want to just pull from that updated file once it is inside of the FC system instead of pulling the info that we need from the object that is made with mWater_field_notes.

Looks good to merge. I think that I am going to work on making this pipeline better in the next few weeks. This will help a ton, I might ping you for mWater questions still though.

src/mWater_collate/grab_sensor_notes.R
SamStruthers: Checked and you are correct.

SamStruthers:
Resolved Juan's comment about not needing to source(`clean_mwater_notes`) in `grab_sensor_notes`

PR #91, opened 2023-12-12
modify slope thresholds

kathryn-willi:
This PR modifies our slope thresholds by first parsing the positive and negative slope values out and grabbing the percentiles from those split-up ranges. I am hopeful this will make the slope function way less greedy, since it will be identifying more extreme values only.


juandlt-csu (approved):
Nice, the only thing that I would add for data that we know to be erroneous is the sensor_spec flag.

kathryn-willi:
Good idea

Issue #73, opened 2023-11-09
Appending 6-hour data from API


kathryn-willi:
Work will be centered within the `api_pull` folder

kathryn-willi:
UPDATE: can we actually do this in 3-hour increments instead?

juandlt-csu:
We can do this in 3-hour increments but we will need to append to the previous 6 hours of data for the large anomaly flag to work on the more recent incoming data point.

juandlt-csu:
This is mostly done, just need to append to historical API data and we should be good. historical flagged data and appending to incoming data is done.

juandlt-csu:
need to revisit this to make sure that the workflow is working how we expect it to when it comes to flagged data since it is being flagged twice.

juandlt-csu:
I am going to mark this as done. Everything seems to be working properly. We have not tested this extensively.

PR #93, opened 2023-12-14
Opening up a PR to merge Juan/PSN/targets to ross/PSN/main

juandlt-csu:
This PR is progress on #73 and updates the flag report analysis.

  1. Updating .gitignore so that it ignores the _targets/ folder (I talked to MB about this)
  2. _targets:
      1. I commented out the appending API data target step because I don’t want it to run yet.
      2. append_historical_api_data.R (special attention to logic here would be appreciated)
          1. This function combines the incoming API data and the historical API data and then deletes the files that are inside of the incoming_api_data folder.
  3. Flag report analysis:
      1. I will replace `explore_and_fix_fxns/plotting_fxns/percent_flagged.R` with `flag_report_analysis2`. 
      2. This revamped function now has
          1. Updated san_na_flag list that is consistent with how we updated the large anomaly flag
          2. A structure that I believe is easier to follow along with more descriptive information on the steps in the function.
          3. This might need to be updated due to the updated slope flag
  4. Updated .Rmd that we were using to be less cluttered
  5. Added add_spec_flag() to make_threshold_table()


src/qaqc/explore_and_fix_fxns/plotting_functions/percent_flagged.R
juandlt-csu, line 1: I started working here but decided that remaking the df would be easier.

_targets.R
```diff
@@ -198,6 +198,7 @@ list(
           add_na_flag() %>%
           add_repeat_flag() %>%
           add_suspect_flag() %>%
+          # we should also be incorporating the add_malfuntion_flag() here no?
```
kathryn-willi, line 201: Still not totally sure.... Ideally, we wouldn't need this function at all, since the malfunction data should be flagged with our other functions..

src/qaqc/download_and_flag_fxns/make_threshold_table.R
```diff
@@ -19,6 +19,8 @@ make_threshold_table <- function(df){
   good_data_stats <- df %>%
     # REMOVE DATA WE KNOW TO BE ERRONEOUS:
     add_malfunction_flag() %>%
+    # we should also add the outside of sensor specification flag here?
+    add_spec_flag() %>%
```
kathryn-willi, line 23: Seems right to me!

_targets.R
```diff
@@ -109,14 +109,14 @@ list(
     rbind(old_field_notes, mWater_field_notes)
```
kathryn-willi, line 109: can we//do we have a dummy mWater_API.yml in the mWater_collate folder with the necessary formatting for it to work?

kathryn-willi (approved):
As it is, I can't run it because I don't think I have the mWater yaml. Once I get one of those - I'll run the pipeline to see if it gives me any errors.

_targets.R
SamStruthers: Question: Do we want to just keep the true mWater yaml in the data folder or in the `mWaterCollate` folder? It has no real sensitive information (api creds, sensor data, etc) just field note data

_targets/meta/meta
```diff
@@ -35,6 +35,7 @@ hv_token|stem|059b668fd0effe34|268e267fa1b9be68|d19cbbef15d7e0e2|268012145||t197
 incoming_data_collated_csvs|stem|526cc5b8a51c611b|c67bb54d6480df1d|46f52353b4288e52|-1436724047||t19703.8407897502s|e1a91a3d0391642b|11588|rds|local|vector|||0.264||
 incoming_data_csvs|stem|0512ae32462262bc|f161ac32f776b1e2|d09e3fc86eeb9554|1255066382||t19689.8810172193s|63e5c65d6d1ce375|134|rds|local|vector|||3656.84||
 incoming_data_csvs_ab32d911|branch||6a31d92726790c7b|c2829ed4f3a9873e|1879218812||t19678.8947293788s||0|rds|local|vector|incoming_data_csvs||0.021||no applicable method for mutate applied to an object of class tryerror
```
SamStruthers, line 37: Q: Should this meta file be in the repo since you ignored the folder earlier?

_targets.R
```diff
@@ -78,6 +78,19 @@ list(
   #   }
   # ),
 
+  # get the test data for each site
+  tar_target(
+    incoming_data_csvs_upload,
+    {
+      start_dt = "2023-11-30 07:00:00"
+      end_dt = "2023-11-30 10:00:00"
```
SamStruthers: Thought: I realize this is to have test data to play with but in practice next year, I am wondering if we should have start_dt be a day or two before sys.date in case we upload old logs (ie vulink not transimitting data but still logging)

SamStruthers (approved):

_targets/meta/meta
juandlt-csu: yes... but since it was already in the repo I will do some fenagling to make it act right. (probably this https://stackoverflow.com/questions/1139762/ignore-files-that-have-already-been-committed-to-a-git-repository)

_targets.R
juandlt-csu: hmm. I think that even in an ideal scenario where these are caught I don't think it would hurt to flag them again?

_targets.R
juandlt-csu: I think we should put it in the mWaterCollate folder so that we stay consistent with what we did for the api_pull. Sam, do you think that we will need to have a readme like what was done for the hydrovu api credentials?

_targets.R
juandlt-csu: How we decide what date times to pull from the hydrovu api will be from some combination of detecting when the last flagged data was and when the last API pull was (right now it is only based on when the last flagged data was). When we upload old logs it will detect when the last flagged date/api pull date was and it will set that as the start date. I think that answers this question?

_targets.R
SamStruthers: If we do, it can be very short essentially explaining what mWater is (link to it), how we use it to record our field notes and then to contact me if they have additional questions or would like to duplicate our field notes. 
The techs did create a table in the SOP with how to fill out each section if we want to include that

PR #94, opened 2023-12-20
 Targets (mostly) working!

kathryn-willi:
Lots of changes! Most are quite minor, but the big things are:

- I created a new Rmd that runs the workflow across the entire 2022-2023 dataset to get the thresholds that are used in the targets workflow. It outputs objects called all_data_flagged_new. 
- I reorganized the _targets.R to better reflect how other targets workflows look in ROSS (namely, listing package dependencies for each tar_target independently).
- Because we are still in development stage, I've got the historical data ending on 11/28 (i.e., the data in all_data_flagged), so that data can be appended for testing the the workflow. I also have hard-coded the end_dt in the api_puller() to end on 11/30 so it doesn't take forever.
- I think I fixed the DT issues. Incoming data should now be getting pulled starting on the last DT of the historic data.
- I think I fixed the field notes munging. The sonde_employed column should now be correct.

If you could run the workflow to see if it works on your end, that would be amazing! Other things that need checked/reviewed:
- Check the column types across old sensor notes and new notes. (For example, sensors_cleaned is a character in the old, and logical in the new.)
- Check time stamps in the all_data_flagged_new.RDS to confirm the DTs are correct like I think they should be. (Best approach I’ve found is comparing values in HydroVu at specific times, and making sure site visits in the field notes align.)
- I'm not totally sure that the network_check function is working correctly. This will require a deep dive at some point.
- I made some changes to how the data is getting appended that may not work in all instances. This will require a deep dive at some point.


SamStruthers:
Pipeline ran for me (once i switched incoming file path to "api/test_data"
Started checking out some of the data (saved to sjs_test_all_data_flagged), seems like some of the mWater field notes are still not being added to the flagged dataset. Also weirdly, no field notes were added to River Bluffs at all....

juandlt-csu (approved):
- I was able to run the targets pipeline.
- The columns across the old field notes and mWater field notes are not aligning when I run str() for both of them.
- I think that the DTs are correct as well.
- I will do a deep dive on the network check
- I will do a deep dive for the appending of data

kathryn-willi:
Sweet! Going to merge, and then delete the data folder so we don't keep encountering those pesky data directory issues.

PR #96, opened 2024-01-16
Fixing small bug in mWater api pull

SamStruthers:
mWater decided to change the allowed forms of datetime which broke our parsing of the datetime columns (start_dt, end_dt and malfunction_end_dt). 
Work: 
- Simple fix in `clean_mwater_notes` to correctly parse DT
- When this was joined with old field notes, end_dt became numeric values. Adding end_dt  = as.POSIXct(NA, tz = "MST")) in `clean_field_notes` seemed to have resolved that and end_dt is a datetime object again

Review Req: Make sure the field notes seem to get pulled in correctly on your side and thoughts on whether we should move to using `parse_datetime` to give multiple formats in the event that this type of change occurs again

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ _targets/
 
 **/images/
 
+#Sam testing script
+scratch/sjs_scratch.Rmd
--- a/scratch/reporting/2023_report_figs.Rmd
+++ b/scratch/reporting/2023_report_figs.Rmd
@@ -11,7 +11,12 @@ lapply(c("tidyverse", "feather", "data.table", "ggpubr", "tidyr", "broom", "flex
 
 #walk(list.files('src/qaqc/explore_and_fix_fxns', pattern = "*.R", full.names = TRUE, recursive = TRUE), source)
 # Source most recent flagging data
-all_data_flagged <- read_feather("data/flagged/all_data_flagged_new.feather") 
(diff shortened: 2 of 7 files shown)


juandlt-csu (approved):
Thanks for fixing this Sam, I will let you know if I get any issues with the mWater data download. 

I think that it would be a good idea to use parse date time to add some robustness to the workflow when it is live.

SamStruthers:
@juandlt-csu I added the datetime parsing if you want to test using that instead of the changes I made before

SamStruthers:
Neglected to merge to main before adding realistic range commits. @kathryn-willi check out the last couple commits for my recent changes. 
Work: 
- created a scratch function `grab_raw_data.R` to quickly pull raw data. I needed to share data with a grad student and now there is a relatively easy way to query what you want from the flagged dataset and the pre 2023 dataset. 
- created `add_realistic_flag.R` and its associated yml. This is essentially a copy of `add_spec_flag` but using realistic ranges rather than sensor ranges. the yml could be trimmed down to only the parameters of interst (ph, sc, temp) but I left everything else from `sensor_spec`
- small updates to figures for 2023, going to chat with MR about it next week to share with FC

kathryn-willi:
Going to merge and start incorporating your new flag as well as a few others. Please stay tuned for another PR from me sometime next week! :)

PR #104, opened 2024-03-04
New {targets}-style mapping approach

kathryn-willi:
Hi Juan,

This PR looks big but it is almost exclusively editorial changes to our comments and some roeorg of some old documents.

The big changes are:

1) Fix to DT parsing in `API_puller.R`. I don't think it was working properly before. This seems to be a recurring issue for us - could you confirm this is working now?
2) Start to "biofilm" flag. This is still not in a great place. I plan to continue working on this for us, so don't worry about reviewing too deeply!
3) I added the `sensor_malfunction.R` flag to our workflow ((finally))!
4) I started adding some {targets} features we weren't using before! Namely, `pattern` which allows the workflow to skip over components of `map` if there is no new data to update within that component, and `require` to track function outcomes that don't return objects. I would like you to continue incorporating these new {targets} features across the entire workflow for us. 

-Katie


_targets.R
```diff
@@ -17,17 +17,15 @@ tar_source(files = "src/mWater_collate")
...
-  ## Incoming API data and related API data archive folder ----
+  # Confirm incoming_api_data folder is empty, which indicates the last
+  # iteration of the data pull + QAQC ran properly.
   tar_target(
     name = verify_incoming_data_dir,
     command = {
       check_incoming_api_dir(incoming_dir = "data/api/incoming_api_data/",
                              archive_dir = "data/api/archive_api_data/")
-      },
+    },
```
juandlt-csu, line 28: question: If we want this to always happen at the beginning of the pipeline, would that mean that we would `require` itself?

_targets.R
```diff
@@ -110,20 +109,18 @@ list(
...
-      },
-    packages = "tidyverse"
+    },
+    packages = c("tidyverse", "data.table")
   ),
 
-  # format data
-
-  ## generate a site parameter combination list
+  # Generate a site parameter combination list to iterate over:
```
juandlt-csu, line 123: praise: I like these more descriptive comments, something I will try to be more cognizant of as well.

_targets.R
```diff
@@ -145,61 +142,82 @@ list(
...
-        set_names(paste0(site_param_combos$sites, "-", site_param_combos$params)) %>%
-        keep(~ !is.null(.))
-    },
+    command = summarize_site_param(site_arg = site_param_combos$sites,
+                                   parameter_arg = site_param_combos$params,
+                                   api_data = incoming_data_collated_csvs,
+                                   notes = field_notes) %>% # to do (j): why do I need to call field_notes here? I don't think we should need to do that...
+      set_names(paste0(site_param_combos$sites, "-", site_param_combos$params)) %>%
+      keep(~ !is.null(.)),
+    pattern = map(site_param_combo), # look up other pattern options, might be other things you can do here
```
juandlt-csu, line 172: question: was this ran and compared to the old output?

the old command calls for a map2 over the columns in site_params_combos, but this uses just the one df. does `pattern = map(site_param_combo` perform this action row wise? I am a little confused how this works.

_targets.R
```diff
@@ -145,61 +142,82 @@ list(
...
     packages = c("tidyverse", "padr")
   ),
 
-  # append to chunk of historical flagged data  ----
-
-  ## get the last 3 hours of the historically flagged data and append it to the incoming data
+  # Get the last 24 hours of the historically flagged data and append it to the incoming data.
```
juandlt-csu, line 176: question: are we appending the last 24 hours now?

_targets.R
```diff
@@ -211,25 +229,37 @@ list(
...
+          # flag for known instances of sensor malfunction:
+          add_malfunction_flag() %>%
+          # store censored, clean data in new column
+          mutate(mean_public = ifelse(is.na(flag), mean, NA))
       })
-      # network check
+
+      # Network check. If certain flags (like a spike in concentrations, super high
+      # values) are also occurring up-/downstream at the same time, unflag it
+      # since it is likely real
```
juandlt-csu, line 253: praise: I still think this is so cool

_targets.R
```diff
@@ -211,25 +229,37 @@ list(
...
+          # if >50% of data is flagged for any reason above in a rolling
+          # 3-hour window, flag all the data:
           add_suspect_flag() %>%
-          # we should also be incorporating the add_malfuntion_flag() here no?
-          mutate(mean_public = ifelse(is.na(flag), mean, NA)) #%>%
-        #mutate(historical_flagged_data_1 = TRUE)
+          # flag for known instances of sensor malfunction:
+          add_malfunction_flag() %>%
+          # store censored, clean data in new column
+          mutate(mean_public = ifelse(is.na(flag), mean, NA))
```
juandlt-csu, line 248: to do: we're going to want to make sure that we are keeping track of how often data is being flagged. I think that's how I determined how to join the data back together after we reuse flagged data to generate the summary stats and reflag the data. If this does not make sense we should talk about it because I have not worked on that in a while and I don't remember if it's actually in place or not.

src/api_pull/api_puller.R
```diff
@@ -25,9 +25,9 @@ api_puller <- function(site, start_dt, end_dt = Sys.time(), api_token, dump_dir)
   # reduce overlapping data
 
   # tz weirdness
-  utc_start_date <- format(as.POSIXct(format(start_dt + hours(7), tz = "MST"), tz = "UTC"), format = "%Y-%m-%d %H:%M:%S")
+  utc_start_date <- format(as.POSIXct(start_dt, tz = "UTC") + hours(7), format = "%Y-%m-%d %H:%M:%S")
```
juandlt-csu, line 28: question: what's going on here?

src/qaqc/download_and_flag_fxns/add_malfunction_flag.R
juandlt-csu, line 1: to do: when sam comes back I want to ask how we can make a more automated system that can generate these flags (maybe by making some sort of sheet in mWater that we read in as well?)

juandlt-csu (approved):
Nice update to the workflow! 

Replies to your list:
1. DT is evil. I will test that it is working but also I think that I want to find where all of our inputs for DT are and figure out why we keep having these issues even though we keep thinking that we fixed the problem.
2. I hope you have a good time working on biofilm. I am not going to focus on that right now, but I am excited to work on it in the future!
3. Nice!
4. I will incorporate `pattern` and `require` where I can!

I also added a couple of comments that I think I want to make issues. Looks good to merge to me!

_targets.R
juandlt-csu: this is essentially making sure that issue #73 is actually completed and working properly.

PR #117, opened 2024-04-22
New intrasensor functions and some half-baked reorg!

kathryn-willi:
Juan,

I started work on some "intrasensor" flagging functions, including a flag looking for times where temp <= 0, and the "biofouling" flag that compares the slope of turbidity in SpC to find drift. 

I also moved the mWater code into the "download_and_flag_fxns". This is something I would like to talk to you about more in-depth, but I think we will need to move/straighten up this repo when we move it into FCW's system. I'm thinking we could move all functions into a single src/ folder, generated tables (like our seasonal thresholds, spec thresholds, etc.) into a thresh/ folder, and any API credentials into a creds/ folder. 

Moreover, I started explicitly calling out packages on all functions. This is something I would like to include across the entire workflow to ensure it's calling the correct thing in FCW's system (basically just a precautionary step).

-Katie


src/qaqc/download_and_flag_fxns/grab_confirmed_mWater_malfunctions.R
juandlt-csu, line 1: Is this where the new sensor malfunction function is being written?

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
```diff
@@ -1,13 +1,20 @@
...
-    # Define a function to check if a given 3-day window's worth of data has a linear
-    # relationship with time
+  # Only test for biofilm growth on turbidity sensors
+  if("Turbidity" %in% df$parameter){
+
+    # subset data to turbidity and conductivity only
+    sub <- df %>%
+      dplyr::filter(parameter %in% c("Turbidity", "Specific Conductivity")) %>%
+      dplyr::select(DT_round, DT_join, parameter, mean) %>%
+      tidyr::pivot_wider(names_from = parameter, values_from = mean)
```
juandlt-csu, line 10: Question: at what point does this function get passed into the pipeline? As I understand the pipeline the dfs go into the flag_all_data() step in site_param form. Is there a step that binds these parameters together that I missed?

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
```diff
@@ -21,37 +28,70 @@ add_biofilm_flag <- function(df){
...
-
-    return(drift_tested)
-
-  } else { return(df) }
-  }
+    # Function to check if a selected time window's mean R-squared value is
+    # at least 60% (i.e., fairly high linear relationship with time indicates a "steady" drift)
+    check_too_steady <- function(x) {
+      mean(x) >= 0.60
+    }
```
juandlt-csu, line 43: Suggestion: I wonder if we would want to statistically verify this R2 value.

Question: Is this checking the R2 value for the 3 hour period that we pulled in?

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
```diff
@@ -21,37 +28,70 @@ add_biofilm_flag <- function(df){
...
+    # for a selected parameter is greater than 60%, we determine it has "failed" (i.e., drift seems to exist)
+
+    biofilm_tester <- function(data = sub, col){
+
+      data %>%
+        data.table::data.table() %>%
+        #bind_rows(flagged_data_dfs[["prospect-Turbidity"]]) %>%
+        # dplyr::filter(lubridate::as_date(DT_round) >= "2022-09-10" & lubridate::as_date(DT_round) <= "2022-10-15") %>%
+        dplyr::mutate(r2_s_right = data.table::frollapply(!!sym(col), n = 96, FUN = progressive_drift, align = "right", fill = NA),
+                      r2_s_center = data.table::frollapply(!!sym(col), n = 96, FUN = progressive_drift, align = "left", fill = NA),
```
juandlt-csu, line 55: question: What does this object name mean? align = "left" but its called center.

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
```diff
@@ -21,37 +28,70 @@ add_biofilm_flag <- function(df){
...
+    }
 
+    biofouling <- c("Turbidity", "Specific.Conductivity") %>%
+      purrr::map(~biofilm_tester(data = sub, col = .)) %>%
+      dplyr::bind_cols() %>%
+      dplyr::rename(DT_join = DT_join...2) %>%
+      dplyr::mutate(parameter = "Turbidity") %>%
+      dplyr::right_join(., df, by = c("DT_join", "parameter")) %>%
+      # If a steady slope is existing through time for turbidity, but NOT for conductivity, it is likely
+      # sensor bio-fouling
```
juandlt-csu, line 71: Suggestion/idea: How long does this function take to run? I don't know how it fits into the workflow right now, and I don't know if I understand the function very well, but I think it might make more sense to have the initial parameter check at the top look for both turbidity and conductivity instead of binding them together before this step. I think that checking if the flags are shared across these parameters could be done in the network_check() function that we have no? 

Though that would mean that we would need to update the network_check() to do intrasensor comparisons and not just intersensor comparisons...

src/qaqc/download_and_flag_fxns/add_frozen_flag.R
```diff
@@ -0,0 +1,20 @@
+add_frozen_flag <- function(df){
+
+  # create a df of temperature for each site
+  temperature <- df %>%
+    data.table::data.table() %>%
+    dplyr::select(DT_round, DT_join, parameter, mean) %>%
+    dplyr::filter(parameter == "Temperature") %>%
+    dplyr::select(DT_join, Temperature = mean)
+
+    # add "temperature" column to df:
```
juandlt-csu, line 10: This could be a good method for adding the other metadata parameters to the data frames that we have.

src/qaqc/download_and_flag_fxns/add_frozen_flag.R
```diff
@@ -0,0 +1,20 @@
...
+    temperature_checked <- df %>%
+      dplyr::left_join(., temperature, by = "DT_join") %>%
+      # If water temperature is freezing, flag all parameters
+      add_flag(., Temperature <= 0 & !grepl("frozen", flag), "frozen") %>%
```
juandlt-csu, line 14: question: I think that we don't have to add the `& !grepl("frozen", flag)` to the flags, that should be handled in the add_flag() function. Have you noticed that it is not working properly?

juandlt-csu (approved):
Hey Katie,

Here are some thoughts in bullets:
- The freeze flagging function seems to be simple enough and gets the job done.
- The biofilm flag is more involved, and I am a bit confused about it. If you are confident that it works within the pipeline I am happy to let it through.
- I think that for the folder re org we might want to pump the brakes on moving stuff right now and instead make a new branch that will be more in line with what we are planning to use in FC system. Let me log in and play around in there before we move on with that plan. 
- I am totally fine with calling functions explicitly. 

Thanks for this update!
Juan

src/qaqc/download_and_flag_fxns/grab_confirmed_mWater_malfunctions.R
kathryn-willi: yep! or an intermediary function. Still TBD

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
kathryn-willi: This happens AFTER the flag_all_data step. See my 00_preparatory_workflow to see where I'm envisioning these interasensor checks happening

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
kathryn-willi: It grabs the average R2 value across both a moving window of 24-hour data and 3-day data . Can definitely be improved!

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
kathryn-willi: That's a mistake - should change to left!

src/qaqc/download_and_flag_fxns/add_biofilm_flag.R
kathryn-willi: I think we want to separate this from our "network_check()" since these comparisons are not happening across the network. Instead they happen at a single site across sensors. But, I like the idea of separating out the stats developed and the flag created from those values!

src/qaqc/download_and_flag_fxns/add_frozen_flag.R
kathryn-willi: Yes!

src/qaqc/download_and_flag_fxns/add_frozen_flag.R
kathryn-willi: I'm not sure... this could be overkill. This IS happening across all flags currently (where we make sure the flag doesn't already exist in the observation). We should double-check this.

PR #120, opened 2024-05-07
Updating documentation for the QAQC functions

juandlt-csu:
Remembering that commits get added onto PRs as they go, so now this is a much bigger PR than originally intended. 

- Commit 1:
*Added documentation to our QAQC functions.* There is still need to add documentation to the rest of the functions that we made. 
- Commit 2:
*Added function to generate site meta-parameters.* The basic idea for this was that I added a step in targets that made a list of meta-parameter data frames for each site that get bound to their respective sites when the data stream goes through the summarize_site_param() function. There were a couple issues with this code that I haven't worked out yet. Namely:
    - Some of the sites did not have all of the meta-parameters (meta-parameters being temperature, battery level, external voltage, and baro)
    - I did not spend any time trying to figure out how to remove these meta-parameters from the rest of the analysis because I didn't know how that would affect downstream stuff (perhaps another reason to start using testthat...)
- Commit 3: 
*Sensor malfunction function updates.* Thanks to Sam and Jed's hard work I was able to automate the malfunction flag. This required a bit of tinkering with the mWater functions and the targets pipeline a bit. The basic idea here is to filter the malfunction records for relevant information and then flag the data based on the intervals that are in the malfunction records. @SamStruthers I changed the mWater functions enough so that they work for what I wanted them to do, but I don't know what kind of effect these changes will have on your workflow with mWater stuff. If this breaks things for you obviously don't approve this and lmk what we can do to fix them.


src/qaqc/download_and_flag_fxns/generate_site_metaparam.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' @examples
+#' generate_site_metaparam(site_arg = "archery", metaparameter_arg = "Battery Level", api_data = incoming_data_collated_csvs)
+#' 
+#' @seealso [summarize_site_param()]
+
+generate_site_metaparam <- function(api_data, require = NULL) {
+
+    sites <- unique(api_data$site)
+    metaparameters <- c("Temperature", "Battery Level", "Baro", "External Voltage")
```
kathryn-willi, line 23: I think I know why some sites don't have metadata - it is because they are AT 500s, so those data are not included in the TROLLs themselves. If we end up requiring this information (aka, not a priority right now) we will need to write some code for pulling additional data from the VuLinks, which are excluded from this QAQC workflow as it is currently.

kathryn-willi (approved):
Great work, Juan! @SamStruthers can you confirm that the changes made to the mWater fxns don't affect how you use them? If that's the case Juan, go ahead and merge ahead of our meeting on Monday. :)

src/qaqc/download_and_flag_fxns/generate_site_metaparam.R
juandlt-csu: That makes sense!

kathryn-willi:
@SamStruthers I went ahead and merged.... I'll test some of your workflows to ensure they aren't broken. And if they are, I'll fix 'em for us!

PR #127, opened 2024-05-14
BIIIIG reorganization

kathryn-willi:
This is the reorganization as we discussed. 

1) @SamStruthers , I've added a new folder called "operations/" where any of your future management work can now live. For example, your current "manager_tools.Rmd". I also created a "timelapse/" folder where all the .Rmd's associated with developing those can now live.

2) @juandlt-csu , all the QAQC workflow now lives in the folder "qaqc_lower/".

2) All R functions now live in a single "src/" folder. I changed all subsequent `source()` script file paths accordingly. (... at least, I really really tried!)

3) All credentials now live in the "creds/" folder and are untracked in .gitignore. I renamed the .yaml's for consistency (**HydroVuCreds.yml** and **mWaterCreds.yml**), and added a template for the mWaterCreds.yaml as a placeholder.

4) I renamed a few of the functions, and changed their calls accordingly. Namely:

- **clean_field_notes()** is now **load_old_field_notes()**
- **clean_mWater_notes()** is now **load_mWater_notesr()**
- **grab_mWater_malfunction_records()** is now **grab_mWater_malfunction_notes()**

I really tried my best to ensure these changes won't break any of the workflows we use a lot. But, this is a big reorg, so it is possible that I missed a couple sneaky source()/function calls. If things start breaking for you, chances are it's this PR's fault!


juandlt-csu (approved):
Nice! Looks good to me. Thanks for doing this Katie :).

PR #129, opened 2024-06-06
Incorporating verification workflow into main repo

juandlt-csu:
Massive PR. Code review is insanely daunting, for a review I recommend cloning my fork and trying to follow along the readme and running through some tests. I will reset the workflow once a review is done, so no worries about breaking anything. If bugs are found, please write them down and send them to me.


kathryn-willi (approved):
Looks amazing! The only things that don't seem to be working are the verification_status is not updating properly. Nor does the skip functionality.

PR #131, opened 2024-06-11
Ironed out sub-daily index bugs

juandlt-csu:
The base R solution to updating the rows in the verified df was not working as intended. The correct row would be replaced, but the information that would be updated in the row was incorrect. Transitioned to a less sleek dplyr method, but it actually works.


kathryn-willi (approved):
Agree it's now a little long-winded, but the important thing is that it now works!

PR #132, opened 2024-06-12
plotting function for sending to collaborators

kathryn-willi:
Here's the start to a function for plotting data at a selected site and set of parameters. Pink points indicate site visits, the light blue line is FC data up in the canyon (Manner's Bridge) and the grey lines are the up-/down-stream locations' data. Temperature, depth, and my pseudo rain data set are always plotted, regardless of the parameter selected. More on the precip proxy... it is literally just the summation of precip events at any FC monitoring site within the sonde's contributing area (i.e., it isn't really rooted in anything scientific/absolute). But! It does _generally_ show when there were rain events happening plus their magnitude and size.

Also, I'm still toying with how we should set the y-lims, since sometimes the data stretches funny (as I know you are aware!). Looking forward to your thoughts and suggestions!

### DIFF
--- a/src/plotter.R
+++ b/src/plotter.R
@@ -1,0 +1,219 @@
+plotter <- function(site_name = "boxelder",
+                    start_date = "2023-03-04",
+                    end_date = "2023-07-07",
+                    title = "wack data title",
+                    parameter = "all") {
+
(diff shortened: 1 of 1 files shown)


juandlt-csu (approved):
Very nice

PR #137, opened 2024-06-27
reverting some mistakes to generate_initial_weekly_plots.R

kathryn-willi:
I reverted those lines we discussed back to their original state. Hopefully this will bring the plots back to their former glory! 

I also added show_col_types = FALSE per @SamStruthers suggestion in add_threshold_lines.R


juandlt-csu (approved):
Thanks for making these updates!

PR #139, opened 2024-07-01
Fixes addressing the FDOM issue

kathryn-willi:
Here's my current approach for solving the FDOM plotting issue. It seemed to occur within the `add_threshold_lines()` function when there are only NAs in a given week of data. So, I just skip that function in those instances. This is a bit of a stop-gap measure; I think the real issue lies in what the function tries to return when there is no data (i.e., I think one of the conditional statements/parentheses are not quite correct). 

I also made a change to the `summarize_site_parameter_full()` function so that field notes are correctly added to Virridy sites (before this PR, they didn't have field notes).

### DIFF
--- a/src/add_threshold_lines.R
+++ b/src/add_threshold_lines.R
@@ -162,7 +162,7 @@ add_threshold_lines <- function(plot, plot_data, site_arg, parameter_arg) {
       return(plot)
     }
 
-  } else if (length(unique_seasons == 1)){
+  } else if (length(unique_seasons) == 1){
 
     site_data <- filter(plot_data, site == site_arg)
 
--- a/src/generate_initial_weekly_plots.R
+++ b/src/generate_initial_weekly_plots.R
@@ -126,17 +126,19 @@ generate_initial_weekly_plots <- function(all_df_list, pending_df_list, site_arg
             labs(x = "Day",
                  y = "Mean")
 
+          if(!all(is.na(week_plot_data$mean))){
           week_plot <- add_threshold_lines(plot = week_plot,
                                            plot_data = week_plot_data,
                                            site_arg = site_arg,
                                            parameter_arg = parameter_arg)
+          }
 
           week_plot <- week_plot +
             theme_bw() +
             scale_x_datetime(date_breaks = "1 day",
                              date_labels = "%b %d",
                              minor_breaks = date_seq,
-                             sec.axis = sec_axis(~., breaks = date_seq, labels = unique(week_plot_data$weekday))) +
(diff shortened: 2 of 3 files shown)


src/generate_initial_weekly_plots.R
```diff
@@ -126,17 +126,19 @@ generate_initial_weekly_plots <- function(all_df_list, pending_df_list, site_arg
...
                                            parameter_arg = parameter_arg)
+          }
 
           week_plot <- week_plot +
             theme_bw() +
             scale_x_datetime(date_breaks = "1 day",
                              date_labels = "%b %d",
                              minor_breaks = date_seq,
-                             sec.axis = sec_axis(~., breaks = date_seq, labels = unique(week_plot_data$weekday))) +
+                             sec.axis = sec_axis(~., breaks = date_seq, labels = 1:length(date_seq))) + #unique(week_plot_data$weekday))) +
```
kathryn-willi, line 141: @juandlt-csu  Can you confirm that this edit of mine is ok?

juandlt-csu: TBD, have to test it out

src/summarize_site_param_full.R
juandlt-csu, line 1: I am assuming these updates are unrelated to the FDOM plotting problem.

src/add_threshold_lines.R
juandlt-csu, line 1: nice catch.

src/summarize_site_param_full.R
kathryn-willi: correct! This change just allows the field notes to be joined to the lower network's virridy sites

src/generate_initial_weekly_plots.R
```diff
@@ -126,17 +126,19 @@ generate_initial_weekly_plots <- function(all_df_list, pending_df_list, site_arg
             labs(x = "Day",
                  y = "Mean")
 
+          if(!all(is.na(week_plot_data$mean))){
```
kathryn-willi, line 129:
```suggestion
         
```

src/generate_initial_weekly_plots.R
```diff
@@ -126,17 +126,19 @@ generate_initial_weekly_plots <- function(all_df_list, pending_df_list, site_arg
...
           week_plot <- add_threshold_lines(plot = week_plot,
                                            plot_data = week_plot_data,
                                            site_arg = site_arg,
                                            parameter_arg = parameter_arg)
+          }
```
kathryn-willi, line 134:
```suggestion
 
```

src/generate_initial_weekly_plots.R
kathryn-willi, line 141:
```suggestion
                             sec.axis = sec_axis(~., breaks = date_seq, labels = unique(week_plot_data$weekday))) +
```

PR #140, opened 2024-07-01
Made shiny app for virridy verification

SamStruthers:
Made a lil shiny app to view virridy data while verifying. Allows user to select site, parameters, and date range for RAW data

No deep review needed since it doesn't affect the workflow at all. I think we could make this alot better later on and I'm defintely open to ideas (log scales, highlighting flags, pulling in raw vs finalized data)

Lmk if i should put this somewhere else, I just wanted to keep it out of the way :)


juandlt-csu (approved):
Very cool, definitely something worth looking into improving for the next year's verification cycles.

kathryn-willi (approved):
We should also move this out of scratch :)

SamStruthers:
Moved app to operations (which is probably more accurate)

PR #142, opened 2024-07-11
Fixing verification workflow errors

juandlt-csu:
Sam has found some edge cases, so here are the updates to fix those. 

- `add_threshold_lines()`
    - Removed the sensor mins from being plotted because they were either all 0 or not useful.
    - Filter for seasonal thresholds first and then check how many seasonal thresholds there are. The logic before would break if there was a season mismatch. This fixed that edge case.
- `generate_initial_weekly_plots()`
    - The important update here is on L105. Arranging by day instead of weekday fixed the inspect some call DT mismatch that Sam noticed, with the side effect that the days on the secondary axis will now sometimes not be 1-7, but some other order. Until this gets fixed, just check the weekday that is associated with your inspect call matches what day you actually want to look at. 
- `2_verification_method`
    - Sometimes the data is in a weird structure that the previous `bind_rows()` call wouldn’t handle.  This if else checks the structure of `update` and cleans it so it’s easier to handle.

Misc
- `plotter()` 
    - These are some changes that I had made when I was using it briefly. This does not need to be included in this commit, so feel free to not approve this.


src/add_threshold_lines.R
```diff
@@ -14,11 +14,11 @@ add_threshold_lines <- function(plot, plot_data, site_arg, parameter_arg) {
   # Determine the unique seasons in plot_data
   unique_seasons <- unique(plot_data$season)
 
-  if (length(unique_seasons) > 1) {
+  # Filter seasonal_thresholds for the seasons present in plot_data
+  seasonal_thresholds <- seasonal_thresholds %>%
```
SamStruthers, line 18: This seems to work correctly for me with PFAL

src/add_threshold_lines.R
```diff
@@ -145,15 +145,14 @@ add_threshold_lines <- function(plot, plot_data, site_arg, parameter_arg) {
...
-        (min(site_data$mean, na.rm = TRUE) <= sensor_thresholds_quantiles[1])) {
-        plot <- plot +
-          geom_hline(aes(yintercept = sensor_thresholds$mix, # *** this needs to be min
-                         color = "Sensor Min",
-                         linetype = "Sensor"))
-      }
-      if (#!is.infinite(max(site_data$mean, na.rm = TRUE)) &
-        (max(site_data$mean, na.rm = TRUE) >= sensor_thresholds_quantiles[2])) {
+      # if ((min(site_data$mean, na.rm = TRUE) <= sensor_thresholds_quantiles[1])) {
+      #   plot <- plot +
```
SamStruthers, line 149: thanks for removing that! it will be really great for looking at lower sites!

src/generate_initial_weekly_plots.R
```diff
@@ -102,7 +102,8 @@ generate_initial_weekly_plots <- function(all_df_list, pending_df_list, site_arg
             # remove empty dfs from the list
             keep(~ nrow(.)>0) %>%
             bind_rows() %>%
-            arrange(weekday)
+            arrange(day)
```
SamStruthers, line 105: inspect some worked correctly for me when testing! strange bug tho

SamStruthers (approved):
Everything worked correctly when I tested it! Thanks for making those changes!

src/generate_initial_weekly_plots.R
kathryn-willi: Yes, a very weird - maybe {lubridate} - bug?

kathryn-willi (approved):
If Sam tested it, I trust it's been fixed! Thanks, Juan

PR #143, opened 2024-07-15
Remove duplicate verification workflow, modified network_check, mod to shiny app

kathryn-willi:
Juan,

 I've deleted the verification steps that exist in our qaqc_lower folder. This is because we will want to verify all the data, _using_ all the data. Currently, all site-parameters now exist in the all_data_flagged_complete.RDS in the virridy folder (i.e., the upper sondes, the lower sondes, and the virridy sondes co-located with the lower sondes). We should rename our "virridy" folders, workflows, etc. to "all_sondes" or something along those lines since it now includes more than just the virridy data. (I'm holding off until the workflow is more stable.)

@SamStruthers this PR additionally includes updates to the shiny app, so that you can log transform the data now.

I've also included a modification to the network check function, which lists the network order appropriately now. It still doesn't have the comparison of co-located data in it, but thought I'd share it with you anyways!

The last change is that I've started adding the function name to all of our functions (see the mWater function modification). I've been working my way through doing this across all functions ahead of putting them in the city's system to reduce potential for changes in function names, etc. if packages add/modify function names in the future!

-Katie


juandlt-csu (approved):
If I understand this correctly, this PR does not have any updates to the verification workflow, it just removes it from qaqc_lower. The other updates are to the network check and sam's shiny app, which are irrelevant to the workflow. If that is correct then this PR is fine to merge by me.

PR #149, opened 2024-07-18
File path call updates

juandlt-csu:
These updates are related to how we call the file paths in the verification workflow to make them more robust across OS's. 
- I use `here()` instead of string concatenation to make those calls more robust across OS's. 
- I replace `readRDS()` with `read_rds()` to make pulling in data more robust across OS's.
- Note that I did not update `saveRDS()` with `write_rds()`. This made the row updates slower and error out. Something to look into in the future. `saveRDS` works fine still.

@kathryn-willi and @SamStruthers -- If y'all could make sure that this doesn't break things on your end and behaves as expected that would be great.

@ecogub -- Nick, if you could test to see if this works at all on your windows machine that would be great. lmk if you can't so I can seek out a window machine to test on.

### DIFF
--- a/virridy/verification_workflow/1_verification_setup.Rmd
+++ b/virridy/verification_workflow/1_verification_setup.Rmd
@@ -8,29 +8,29 @@ library(tidyverse)
 library(ggpubr)
 library(here)
 
-# set data directory paths
-all_path <- here("data", "virridy_verification", "all_data_directory")
-pre_verification_path <- here("data", "virridy_verification", "pre_verification_directory")
-intermediary_path <- here("data", "virridy_verification", "intermediary_directory")
-verified_path <- here("data", "virridy_verification", "verified_directory")
-
 # set function directory paths
-fxn_path <- "src/"
+fxn_path <- here("src")
 
 # Source functions
 invisible(map(
   list.files(fxn_path, pattern = "*.R", full.names = TRUE),
   ~ source(.x, echo = FALSE)
 ))
+
+# set data directory paths
+all_path <- here("data", "virridy_verification", "all_data_directory")
+pre_verification_path <- here("data", "virridy_verification", "pre_verification_directory")
+intermediary_path <- here("data", "virridy_verification", "intermediary_directory")
+verified_path <- here("data", "virridy_verification", "verified_directory")
   
 # Load all flagged data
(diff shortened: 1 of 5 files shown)


virridy/verification_workflow/1_verification_setup.Rmd
```diff
@@ -42,7 +42,7 @@ get_working_data_decision()
 network <- "virridy"
 # Set the site and parameter to verify
 site <- "archery"
-parameter <- "Temperature"
+parameter <- "Chl-a Fluorescence"
 ```
 
 ```{r set the site-param df to update}
```
ecogub, line 48: Throwing error from chunk 5 "Error in UseMethod("group_by") :
no applicable method for 'group_by' applied to an object of class "NULL""

virridy/verification_workflow/1_verification_setup.Rmd
```diff
@@ -71,7 +71,4 @@ weekly_plot_objects <- generate_initial_weekly_plots(
   flag_arg = NULL
```
ecogub, line 71: Same error as chunk 5 is thrown here in chunk 6.

virridy/verification_workflow/2_verification_method.R
ecogub, line 1: Running this with the errors above gives the error: "Error in weekly_plot_objects : object 'weekly_plot_objects' not found"

ecogub (changes requested):
The things I pointed out in the meeting run fine, just the last 2 chunks in step 1 are throwing errors.

virridy/verification_workflow/1_verification_setup.Rmd
juandlt-csu: Chunk 5 (set the site-param df to update) is giving you an error that has "group_by" in it?

virridy/verification_workflow/1_verification_setup.Rmd
juandlt-csu: A group_by error makes more sense here, unfortunately I can't debug without a reprex. Thanks for this, I will seek out a windows machine to figure this one out.

virridy/verification_workflow/2_verification_method.R
juandlt-csu: This makes sense. if you were unable to run chunk 6 then you wouldn't have the list of plots needed for this script. Thanks for checking this out, Nick!

PR #151, opened 2024-08-02
overhaul of flagging workflow!!

kathryn-willi:
This is a massive PR! But, all is centered around the new workflow for flagging found in `virridy/non_targets_workflow/0_auto_qaqc.Rmd`. I suggest reviewing this workflow with a fine comb, as the steps in qaqc pipeline have changed quite drastically. 

Namely, I had to bake in the option of performing this workflow across either ALL sondes, or just those in the FCW network, which led to a lot of changes in the majority of our older functions. I also started to explore pulling in data that's been verified to make better flags (see the "define thresholds" section in the Rmd). And, you will also see that there are new flags I've developed including a way to flag DO "noise", capping turbidity at 1000 NTU, and a bad-calibration fixer that is still a work in progress. 

There is also a weird thing that needs to be done with how we first pull in our API data (data/api/krw_everything_backup2.0_mdt_dl/). This is because data from 2023 can no longer be pulled from HydroVu (names have changed). So with that, **please do not attempt to download any API data before 2024!** I had to write code to "munge" the old data correctly, `munge_api_data_hacky()`, that we should remove entirely once we move past working with older data.

Something else to be mindful of is the idea that this workflow works across all data in one go. We will need to modify `0_auto_qaqc.Rmd` and likely some of the functions to make them work when we migrate to doing this process in real-time/tacking on new data with the old data.


virridy/non_targets_workflow/00_virridy_preparatory_workflow_old.Rmd
```diff
@@ -154,10 +159,6 @@ threshold_lookup <- all_data_summary_stats_list %>%
   dplyr::bind_rows()
 
 readr::write_csv(threshold_lookup, 'data/qaqc/seasonal_thresholds_virridy.csv')
-
-# also save the threshold lookup table as a RDS 
-# saveRDS(threshold_lookup, 'data/qaqc/seasonal_thresholds_virridy.RDS')
-
 ```
 
 *Compare seasonal thresholds to our own "reasonable" ranges for parameters generally:*
```
juandlt-csu, line 164: Question: This does not need to be part of the FC system right?

virridy/non_targets_workflow/0_auto_qaqc.Rmd
juandlt-csu, line 1: Question: Is this the workflow that I want to incorporate into the FC system? As opposed to the other .Rmd that ends with `_old`.

virridy/non_targets_workflow/00_virridy_preparatory_workflow_old.Rmd
juandlt-csu, line 1: Question: This is NOT the workflow that I incorporating into the FC system, right?

src/add_drift_flag_old.R
juandlt-csu, line 1: Question: Can this function be moved to the scratch folder?

src/add_malfunction_flag.R
```diff
@@ -27,24 +19,84 @@ add_malfunction_flag <- function(df, malfunction_records){
...
+
+    try(df <- df %>%
+          add_flag(DT_round %within% depth_interval_list, "depth calibration malfunction"))
+
+    try(df <- df %>%
+          add_flag(DT_round %within% unsubmerged_interval_list, "sonde unsubmerged"))
+
+    try(df <- df %>%
+          add_flag(DT_round %within% general_interval_list, "sensor malfunction"))
+
```
juandlt-csu, line 96: Question: Does this error out unless its wrapped in `try()`?

src/api_puller.R
```diff
@@ -1,70 +1,113 @@
...
+  if(network %in% c("all", "All", "virridy", "Virridy")){
+
+    locs <- hv_locations_all(hv_token)
+
+  } else if(network %in% c("csu", "fcw", "CSU", "FCW")){
+
+    locs <- hv_locations_all(hv_token) %>%
+      dplyr::filter(!grepl("virridy", name, ignore.case = TRUE))
+
+  }
```
juandlt-csu, line 22: Question: FC system will only have CSU/csu/FCW/fcw networks?

src/fix_calibration.R
```diff
@@ -0,0 +1,389 @@
+#' @title Correct data with underlying bad calibrations using In-Situ calibration reports
+#'
+#' @description This function uses calibration reports collected by field crews to "correct"
+#' instances of poor calibration. These are infrequent, and this function is not yet complete.
```
juandlt-csu, line 4: Question: This says the function is not yet complete, is it incorporated into the pipeline? I think that a portion of the `add_malfunction_flag.R` references this function and says that its made redundant by this function.

src/fix_calibration.R
```diff
@@ -0,0 +1,389 @@
...
+#' @param df Site-parameter data frame
+#' @param cal_errors A data frame containing a list of known instances of poor calibration
+#'
+#' @return A data frame with modified `mean` values corrected for improper calibration. These observations are also identified in the `cal_fix` column.
+#'
+#' @examples
+#' add_range_flags(df = all_data_flagged$`archery-Actual Conductivity`)
+#' add_range_flags(df = all_data_flagged$`boxelder-Temperature`)
+
+fix_calibration <- function(df, cal_errors = readxl::read_excel("data/calibration_error_log.xlsx")){
```
juandlt-csu, line 15: Note: I am realizing that we have a couple of look up tables that are being updated manually still that the FC system will need. Tagging this so that we think of a system to update those easily within the FC system.

src/fix_calibration.R
juandlt-csu, line 1: Note: this seems like it needs to be automated like how we automated the malfunction function.

virridy/non_targets_workflow/0_auto_qaqc.Rmd
```diff
@@ -0,0 +1,452 @@
...
+                             summarize_interval = "15 minutes") %>% #other example: "1 hour"
+    dplyr::filter(lubridate::year(DT_round) == 2023)
+}
+
+if(hacky == TRUE){
+  
+  all_data <- munge_api_data_hacky() %>%
+    dplyr::filter(lubridate::year(DT_round) == 2023)
+  
+}
```
juandlt-csu, line 107: Question: what determines if a pull is hacky or not?

juandlt-csu (approved):
This looks good, ready to merge and convert it to the FC system

virridy/non_targets_workflow/00_virridy_preparatory_workflow_old.Rmd
kathryn-willi: correct! this will not be added to FC

virridy/non_targets_workflow/00_virridy_preparatory_workflow_old.Rmd
kathryn-willi: right, this not the workflow that will go into FC

src/add_drift_flag_old.R
kathryn-willi: yes!

src/add_malfunction_flag.R
kathryn-willi: It must have, or I wouldn't have wrapped them in `try()` ... but I honestly don't remember what causes the error. I think it errors if the df doesn't have anything listed in the `_iterval_list` objects...

src/api_puller.R
kathryn-willi: Yep! All the Virridy stuff can and should get removed from this.

PR #152, opened 2024-08-27
Simple fixes for verification tool

juandlt-csu:
Sometimes the data gets transformed into things that are not what they came in as (come in as df, come out as list) which causes errors when we are updating the data. Those fixes in the verification workflow and function pertain to that.

### DIFF
--- a/src/add_threshold_lines.R
+++ b/src/add_threshold_lines.R
@@ -8,7 +8,7 @@ add_threshold_lines <- function(plot, plot_data, site_arg, parameter_arg) {
     bind_rows()
 
   seasonal_thresholds <- read_csv(here('data', 'qaqc', 'seasonal_thresholds_virridy.csv'), show_col_types = FALSE) %>%
-    #to do: Check to make sure seasonal thresholds csv is not necessary
+    # to do: Check to make sure seasonal thresholds csv is not necessary
     #bind_rows(read_csv(here('data', 'qaqc', 'seasonal_thresholds.csv'), show_col_type = FALSE),
     distinct(site, parameter, season, .keep_all = TRUE) %>%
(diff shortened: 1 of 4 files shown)


src/add_threshold_lines.R
```diff
@@ -7,8 +7,8 @@ add_threshold_lines <- function(plot, plot_data, site_arg, parameter_arg) {
   sensor_thresholds <- yaml::read_yaml(here("data", "qaqc", "sensor_spec_thresholds.yml"))[[parameter_arg]] %>%
     bind_rows()
 
-  seasonal_thresholds <- read_csv(here('data', 'qaqc', 'seasonal_thresholds_virridy.csv'), show_col_types = FALSE) %>%
-    #to do: Check to make sure seasonal thresholds csv is not necessary
+  seasonal_thresholds <- read_csv(here('data', 'qaqc', 'seasonal_thresholds.csv'), show_col_types = FALSE) %>%
```
kathryn-willi, line 10: good catch!

src/add_threshold_lines.R
```diff
@@ -176,14 +176,14 @@ add_threshold_lines <- function(plot, plot_data, site_arg, parameter_arg) {
       # seasonal thresholds
       seasonal_thresholds_quantiles <- unname(quantile(c(seasonal_thresholds$t_mean01, seasonal_thresholds$t_mean99), c(0.1, 0.9)))
 
-      if ((min(site_data$mean, na.rm = TRUE) <= seasonal_thresholds_quantiles[1]) == TRUE){
```
kathryn-willi, line 179: ooo another good catch!

kathryn-willi (approved):
Fumigating those bugs!

PR #154, opened 2024-09-23
Fixing issues/ improving raw data plotter

SamStruthers:
Juan had run into some issues with the raw data plotter which seemed to be related to the generation of base_plot. These should resolved and allow users to plot single or multiple sites. 

I also made the plot larger and gave users the ability to hide the selector page so they can only look at the plot. 

I also improved some of the coloring of the sites/flags. Sites now have a specific color scale so they will always be the same color when swapping between sites/parameters. I would like to do a similar thing with flags but when I was testing it, the colors were very hard to distinguish since there are ~150 unique combos of flags

Review: Just testing the plotter to make sure it works on your end and if there are any other improvements to make :)


juandlt-csu (approved):
Thanks for making these updates, Sam!

PR #158, opened 2025-01-30
Shiny Ver Tool V1

SamStruthers:
WOOO! A semi working app!

All code is located in the directory: `shiny_ver_tool/ver_tool_v1`
@juandlt-csu  I emailed you the data folder that needs to be in this folder for the app.R file to work for testing. 

Work:
- Data selection tab works. Allows user to select initials, directory, site, param and sub plot params (auto filled from meta csv file)
- When load data button is pressed, user is sent to data verification tab and data for that site-param is loaded into `selected_data()`
- User can move between weeks and back to the data selection tab if needed
- Users can brush points, if brush points enabled. This can change the flag and omit columns in selected_data()
- Users can make weekly decisions. They are shown a preview of the final dataset based on their decision and can hide points that would be omitted by their decision. If they make a weekly decision, they will be moved to the next week and selected_data() columns (flag, omit, final_status, user) are updated based on the decision and existing flags/omits
- Users can also toggle the subplots and are shown the data from other parameters selected in data selection page

Review/moving forward:
- Make sure it works on your end
- I added to do's or Qs throughout the script so thoughts or feedback would be great there
- I think top priority is getting the data to be saved in the int directory and overwriting/saving at decision points (Weekly decision, quit, etc)
- We should also make sure that if there is verified data from other sites/parameters, we want to pull that data into the main/sub plots whenever possible
- Improving the weekly plot to include other sites, seasonal ranges and one/two additional days of data on either end would be great
- With a smaller dataset it should be easier to test getting all data verified and move to the final tab to finalize dataset.

### DIFF
--- a/shiny_ver_tool/ver_tool_v1/app.R
+++ b/shiny_ver_tool/ver_tool_v1/app.R
@@ -213,7 +213,7 @@ server <- function(input, output, session) {
 #### Reactive values ####
   data <- reactiveVal(NULL)
   current_week <- reactiveVal(NULL)
-  selected_data <- reactiveVal(NULL)
+  selected_data <- reactiveVal(NULL) # site param df
   all_datasets <- reactiveVal(NULL)
   brush_active <- reactiveVal(FALSE)
 
@@ -545,6 +545,8 @@ server <- function(input, output, session) {
             between(mean, brush_mean_min, brush_mean_max) & user_brush_select == "A" ~ as.character(NA),
             #Flag
             between(DT_round, brush_dt_min, brush_dt_max) &
+
+  #TO DO: Turn into function add flag
             between(mean, brush_mean_min, brush_mean_max) &  user_brush_select == "F" ~ as.character(flag_choices),
             #Omit
 #TO DO: If a user selects Omit, do they need to give the data a flag?
@@ -554,9 +556,12 @@ server <- function(input, output, session) {
             TRUE ~ flag),
 
           #if a user brushes points as omit, then change omit to TRUE
-          omit = case_when(
+        omit = case_when(
             between(DT_round, brush_dt_min, brush_dt_max) &
             between(mean, brush_mean_min, brush_mean_max) &  user_brush_select == "O" ~ TRUE,
+            #Accept
+            between(DT_round, brush_dt_min, brush_dt_max) &
+              between(mean, brush_mean_min, brush_mean_max) & user_brush_select %in% c("A","F")  ~ NA,
             TRUE ~ omit),
           #if a user brushes points, add their initials to the user column
           user = ifelse(between(DT_round, brush_dt_min, brush_dt_max) &
--- a/shiny_ver_tool/ver_tool_v1/app.R
+++ b/shiny_ver_tool/ver_tool_v1/app.R
@@ -6,6 +6,7 @@ library(lubridate)
 library(here)
 library(ggpubr)
 library(gridExtra)
+library(plotly)
 
 ##### Helper functions for data loading #####
 load_data_directories <- function() {
@@ -77,7 +78,7 @@ get_auto_parameters <- function(parameter) {
   }, error = function(e) character(0))
 }
 
-final_decision_colors <- c("PASS" = "green",
+final_status_colors <- c("PASS" = "green",
                            "TAG" = "yellow",
                            "OMIT" = "red")
 # All available parameters for sub-parameter selection
@@ -199,17 +200,43 @@ ui <- page_navbar(
 
 #### Tab 3: Final Data View ####
 ## To do: This doesn't work at all yet, just a placeholder
-  nav_panel(
-    title = "Final Data",
+nav_panel(
+  title = "Finalize Data",
+  layout_columns(
+    col_widths = c(8, 4),
+
+    # Main Plot Card
     card(
-      card_header("Processed Dataset"),
+      card_header("Final Data Overview"),
       card_body(
-        downloadButton("download_data", "Download Data"),
-        br(), br(),
-        DTOutput("final_table")
+        plotOutput("final_plot", height = "400px")
+      )
+    ),
+
+    # Week Selection and Actions Card
+    card(
+      card_header("Modify Verification"),
+      card_body(
+        selectInput("final_week_selection", "Select Week:", choices = NULL),
+        actionButton("goto_final_week", "Return to Selected Week",
+                     class = "btn-primary w-100 mb-3"),
+        hr(),
+        actionButton("submit_final", "Submit Final Changes",
+                     class = "btn-success w-100")
+      ),
+      card_footer(
+        div(
+          class = "d-flex justify-content-between",
+          div(
+            checkboxInput("remove_omit_finalplot", "Remove omitted data from plot", value = FALSE),
+            style = "margin: auto"
+          )
+        )
       )
     )
   )
+)
+
 )
 
 
@@ -383,7 +410,7 @@ server <- function(input, output, session) {
           title = paste0("Weekly Data for:", input$site, "-", input$parameter),
           x = "Date",
           y = input$parameter )+
-        scale_color_manual(values = final_decision_colors)
+        scale_color_manual(values = final_status_colors)
 
       plot(p)
     } else {
@@ -649,13 +676,14 @@ server <- function(input, output, session) {
     idx <- which(weeks == current)
 
     # Move to next week if available
-    if (idx < length(weeks)) {
-      current_week(weeks[idx + 1])
-    }else{
-#TO DO: This should actually check if all data has been reviewed and move to final tab if all data has been verified
+    if(all(!is.na(selected_data()$final_status))){
+
       showNotification("All weeks have been reviewed.", type = "message")
-      updateTabsetPanel(session, inputId = "tabs", selected = "Data Verification")
+      updateTabsetPanel(session, inputId = "tabs", selected = "Finalize Data")
+      } else{
+      current_week(weeks[idx + 1])
     }
+#To Do: If next week has been reviewed, move to closest week without verified data
 
     # Show notification of submission
     showNotification(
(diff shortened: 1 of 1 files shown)


SamStruthers:
Got inspired this AM (got that feature fever haha), and finished v1 of the final tab
- When all data is verified (all final status column is NA), you are moved to the finalize data tab and shown all the data with the colors from final_status_colors. 
- User can then go and pick a week (based on the numbers at the top) and return to verification tab when they submit changes for the week they are returned to the final data tab

Review/to do:
- Ideally this final plot would be a plotly object but it wasn't playing nice with the weekly vlines
- Right now the data verification tab just moves to the next week unless all data has been verified, maybe we want to change it so that it moves to the most recent week of unverified data rather than just the next week automatically?

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+library(gridExtra)
+library(plotly)
+
+##### Helper functions for data loading #####
+load_data_directories <- function() {
+  list(
+    all_path = "data/all_data_directory",
+    pre_verification_path = "data/pre_verification_directory",
+    intermediary_path = "data/intermediary_directory",
+    verified_path = "data/verified_directory")
```
juandlt-csu, line 17: To Do: Convert to {here} paths

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+# All available parameters for sub-parameter selection
+available_parameters <- c("Specific Conductivity", "Temperature", "pH",
+                          "Turbidity", "DO", "Depth")
+#TO DO: This only has the primary datasets for testing purposes need to add (FDOM, CHLA, ORP)
+###### End Helper Functions ######
+
+
+# UI Definition
+ui <- page_navbar(
+#To Do: remove header to save space? it shouldnt need to be used by users
```
juandlt-csu, line 93: Maybe an option to toggle it away would be better, in the public version. Internal version It could probably go away. I think it should be pretty easy to just toggle it away though.

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+
+      # Sub Plots Card (conditionally shown)
+      uiOutput("subplot_card"),
+      uiOutput("brush_card")
+      )
+    )
+  ),
+
+#### Tab 3: Final Data View ####
+## To do: This doesn't work at all yet, just a placeholder
```
juandlt-csu, line 202: I thought this was fixed in the last commit?

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+        mutate(omit = NA,
+               user = NA,
+               final_status = NA)
+
+
+      # Store the processed data
+      selected_data(processed)
+
+      # Set initial week
+      #To Do: This should update to first week with unverified data if possible
```
juandlt-csu, line 326: This is done in the manual verification by checking the week for any SKIPS in the `verification_status` column or for any NA's in the `is_verified` column

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+      )
+    )
+  })
+
+## Sub plots output
+  output$sub_plots <- renderPlot({
+    req(all_datasets(), current_week(), input$show_subplots, input$sub_parameters)
+# See notes in Card UI Page
+#To do: this should probably be changed to make it faster?
+#Q: can most of the code from raw data plotter be migrated over?
```
juandlt-csu, line 471: I'm sure that a lot of the logic and code can be, but I think that a lot of the interactions and connections between these components are very different so those things need to be understood at least a little bit before a great Copy Paste Migration into this system.

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+  observeEvent(input$submit_final, {
+    showNotification("Final changes submitted successfully!", type = "message")
+    updateNavbarPage(session, "navbar", selected = "Data Selection")
+  })
+
+  #### Extras ####
+  # Handle quit button
+  observeEvent(input$quit_app, {
+    stopApp()
+  #TO DO: Update backend and save file
```
juandlt-csu, line 773: That's me :)

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+    updateTabsetPanel(session, inputId = "tabs", selected = "Data Verification")
+
+  })
+
+  #### Data Verification functions ####
+# Previous Tab
+  observeEvent(input$prev_tab, {
+    updateNavbarPage(session, "tabs", selected = "Data Selection")
+
+#Q: Should this update the data files or no?
```
juandlt-csu, line 346: I think this should update the intermediary data files, no?

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+## Main plot
+  output$main_plot <- renderPlot({
+    req(selected_data(), current_week())
+
+    week_data <- selected_data() %>%
+      filter(week == current_week())
+
+    # Check the decision and create appropriate plot
+    if (input$weekly_decision != "s") {
+#Q: not sure if this is necessary?
```
juandlt-csu, line 379: I am not sure what this question is referencing. If it is referencing making an object that holds user input, I think that this is good, and makes the user inputs easier to track.

shiny_ver_tool/ver_tool_v1/app.R
```diff
@@ -0,0 +1,778 @@
...
+     #Add Omitted data in red
+      geom_point(data = week_data %>%filter(omit == TRUE),aes(y = mean), color = "red")+
+      labs(
+        title = paste0("Weekly Data for:", input$site, "-", input$parameter),
+        x = "Date",
+        y = input$parameter )
+
+   # Add brush rectangle if brush is active and brush exists
+
+#Q This might need to be reactive later on ?
```
juandlt-csu, line 429: Making the brush rectangle reactive is a wish list of mine, though I don't know how to do it yet.

juandlt-csu (approved):
Hey Sam, This is some great work. I commented to answer questions and my thoughts on the to dos where I felt relevant. Most of the to dos seem like details in the front end that need to get worked out, but the meat and potatoes of the front end here looks really great. I want to respond to the review points here:

- Make sure it works on your end: Works awesome :)
- I added to do's or Qs throughout the script so thoughts or feedback would be great there: added feedback where I though necessary or useful.
- I think top priority is getting the data to be saved in the int directory and overwriting/saving at decision points (Weekly decision, quit, etc): This is some back end stuff that I will work out. I want to use RSQLite to make the babiest db that we can use to formalize some file structure here. I think it would be useful. 
- We should also make sure that if there is verified data from other sites/parameters, we want to pull that data into the main/sub plots whenever possible: I think we did this in the manual verification tool, I will need to look for those functions. 
- Improving the weekly plot to include other sites, seasonal ranges and one/two additional days of data on either end would be great: This is also a migration from the old verification tool. Unfortunately I remember making these plots was not super straightforward, but there's lots of code that we can use already. 
- With a smaller dataset it should be easier to test getting all data verified and move to the final tab to finalize dataset.: Are we using a smaller dataset for testing already? I thought that we had some test data in the data folder.

This is awesome dude, excited to see how this turns out.

shiny_ver_tool/ver_tool_v1/app.R
SamStruthers: Gotcha, added a comment on line 319 so we can add in those columns (verification_status, is_verified column), unless that should be done in a pre processing step (ie going from all data directory -> pre_ver directory)

shiny_ver_tool/ver_tool_v1/app.R
SamStruthers: I can play with that independently while you are getting the backend stuff going and see how nicely it wants to play

shiny_ver_tool/ver_tool_v1/app.R
SamStruthers: I think this was just me wondering at what points do we want to save user progress. Obviously we want to save when a user makes a weekly decision but do we also want to save whenever they use the brush or go back to data selection? I think not but perhaps it would be helpful for the user?

shiny_ver_tool/ver_tool_v1/app.R
SamStruthers: Yeah this was referencing, weekly_decision <- input$weekly_decision
Mostly just me being unfamiliar with shiny workflows :)

SamStruthers:
Thanks for the feedback. I replied to the relevant comment and altered a couple comments in the script itself
I'll let you have at the backend side of things, excited for it to do all the stuff!
FYI, I didn't realize there are a ton of files in the `all_data_directory` test dataset I sent you, if you don't need them for plotting, the app is SO much faster with them removed for testing purposes. 

I may have a bit of time this week so I might play around with trying to insert raw data plotter into the subplot card but we shall see if I actually have time

shiny_ver_tool/ver_tool_v1/app.R
juandlt-csu: In the previous version I added these columns during the set up step, but I think it makes more sense to add those columns to the data as needed for this system. 

It seems like the data is loaded into the app via file paths, and it would be nice to have these alterations to the data frames done in the app while the user is making decisions so that the user doesn't have to add more columns to their data frame while they are setting it up.

shiny_ver_tool/ver_tool_v1/app.R
juandlt-csu: I think it makes a lot of sense for the intermediary data to be updated and saved only when a user confirms a weekly decision. I agree that we should not save the data if the user brushes the data and then decides to move on to the next or previous week.

juandlt-csu:
> Thanks for the feedback. I replied to the relevant comment and altered a couple comments in the script itself I'll let you have at the backend side of things, excited for it to do all the stuff! FYI, I didn't realize there are a ton of files in the `all_data_directory` test dataset I sent you, if you don't need them for plotting, the app is SO much faster with them removed for testing purposes.
> 
> I may have a bit of time this week so I might play around with trying to insert raw data plotter into the subplot card but we shall see if I actually have time

It might be worth it to try to figure out some caching stuff for data so that it's not super slow when we are actually wanting to use it.

SamStruthers:
So true! I think we can probably change the tab 1 functions (`load_all_datasets`, `get_sites`, `get_parameters`) and  to only look at file names rather than just loading All the data and then only load data as necessary for the sub plots or when setting up `selected_data()`

PR #159, opened 2025-02-11
Backend synchronization

juandlt-csu:
Hey Sam,

Sorry for this massive PR. Most of the stuff in here is actually not relevant to the PR (it's mostly moving functions (with documentation) from the pipeline into a separate folder for the yet to be completed package). The most relevant files are:

1. `shiny_ver_tool/ver_tool_v1/R/data_synchronization.R`
    - I made a folder for functions specific to the tool. Right now it just has the one file. This file has the new and improved file checking for the back end files. I think it's a lot more robust. You'll notice that at the bottom of this script I have the "move data around" functions that are still in pseudocode mode. I haven't attached them to the front end because I haven't fully figured them out yet and I haven't fully considered the front end specifics for how the app handles data and directory information. Look over these functions and see if you can poke some holes in them. Documentation is lacking, sorry if it's not super clear. Push it back if you want some more documentation.
2. `shiny_ver_tool/ver_tool_v1/app.R`
    - Very slight modifications here. Just some updates to file paths for non-mac users, some libraries for the back end stuff, and a suggestion for a to do. 

Also included in the scratch folders is some of the work that I did trying to make the database idea work. Fun to poke around in, but it's mostly bones that I was planning around until I realized that it wouldn't work.


shiny_ver_tool/ver_tool_v1/R/data_synchronization.R
```diff
@@ -0,0 +1,380 @@
+sync_file_system <- function() {
+
+  pre_dir_path <- here("shiny_ver_tool", "ver_tool_v1", "data", "scratch_data")# pre_verification_path
+  int_dir_path <- here("shiny_ver_tool", "ver_tool_v1", "data", "scratch_data")# intermediary_path
+  ver_dir_path <- here("shiny_ver_tool", "ver_tool_v1", "data", "scratch_data")# verified_path
```
SamStruthers, line 5: I'm assuming this would be changed to the actual file paths later on?

SamStruthers:
For some reason this was not working when I just called `sync_file_system()` but seemed to work when I ran the lines to get file directories and filenames then ran the whole function. Otherwise it seemed to be working? It couldn't open the files (see screenshot), which I am assuming that is bc it is test data but if that is not happening for you, it might be something to look into 
Screenshot 2025-02-11 at 3 45 02 PM

Looking over the pseudocode, I think that all makes sense to me. I left one comment there but I realize a lot of that code is not not written yet. Maybe the next step is to make a new test_data folder with the datetimes/hashes, the actual folder structure we want and data  that should load properly(if that was happening to you too) and maybe we can start seeing how it would update when in the shiny app?

PR #160, opened 2025-02-25
Shiny Ver Tool Complete! (I think?)

SamStruthers:
I think the ver tool is complete in terms of the functionality we set out in our wishlist!

Work completed:
- Brush multiple areas: 
You can now make multiple bounding boxes and submit the same decision for all of them
- Additional sites in main plot: You can now select additional sites to be plotted as lines in the main plot
- Threshold lines: User can toggle on/off the seasonal threshold lines
- Hide Omit/Flagged data:These data (user selected or auto flags) can be omitted, which is nice for removing outliers
- Two days on either end: Users can add/remove two days from either end of the plot for additional context
- Reset week button:  Resets week to default data
-Subplots: Pulled some code over from the original tool and it should correctly select the data from the most verified directory.
- Final Tab: Final plot is now interactive and users see the week # to go back. Much nicer when looking at longer timeseriesSimilarly, can remove omitted data from this. Submit Final Changes brings users back to load data tab. 

No huge rush on this @juandlt-csu, I feel like this is reaching a stopping point and now we just need to work on the file saving side of things. If there is something you want me to take a crack at while you're busier, I'm happy to give it a shot but you probably have a vision for this. Just lmk! 

A couple of general TODOs:
- Grab the correct column (mean, mean_verified, etc) for additional site's data in subplot/mainplot. I wasn't sure how this would end up in final datasets so figured we could hold off till then.
- Find a more efficient way of loading data. As the data folder gets bigger (and we run on tech's computers), it would be good to try to reduce how many site/param combos we try to load at each step. I think this will be easier once we figure out the file saving side of things.


SamStruthers:
Hi @juandlt-csu 
This has become a bit larger of a PR but I think I was able to successfully use the code you wrote in `data_synchronization` to successfully move/update files in the backend! I have moved the data to a folder in the psn folder, `shiny_app_ver/2024_cycle/data` and created a symlink from there to the `shiny_ver_tool` folder. 

Updates:
- I moved everything out of the ver_tool_v1 folder and should have fixed all the sourcing issues. 
- The first tab now includes a reference for techs as to where each site/param is located. 
- The sub plot/main plot exclude omitted data (still showing flagged/passed data), which should improve the usability of other sites as they get cleaned up. 
- The selected dataset is updated at the following points:
-- Upon selecting/loading dataset
-- Upon making a brush decision
-- Upon making a weekly decision
-- Upon quitting/returning to selection page
The dataset will no be moved to the verified directory until the user clicks the "Submit Finalized Dataset"

Question for you: 
The code currently keeps three copies of each intermediary dataset and then adds the excess to the archive folder. I'm wondering if this is necessary due to a tech's ability to go back to a week and undo their choices? It is leading to a very full `int_archive` folder...

When you have a chance, test it out! I played with it a bit and it seems to correctly move files/react normally to the actions in the shiny app. Lmk if you run into any bugs! If I have time tomorrow, I'll probably go through and try to clean up the code as there are relics of me testing scattered throughout...

juandlt-csu (approved):
This looks so great! Sorry for the super, super late review. 

Regarding your question I think there are three options that make sense, depending on what we want to use the archive folder for:

If we would at any point want to review that archive data we can just keep it. 

If we don't need to keep the data for any reason aside from provenance, we could just do away with the archive folder and not keep that data at all. The archive folder is not necessary for roll back and I don't think that there are other functions that check that directory for consistency anyway.

If doing away with the archive data completely is scary, then we can just keep the latest 3 files like the int directory.

Thanks for the patience, Sam. Excited to start using this thing!

SamStruthers:
Removed int_archive as Juan and I decided it is not necessary if a user can go through and reset weeks on their own. 
Added log 10 functionality for plot (useful for turbidity)
Fixed week line issues on final plot

PR #163, opened 2025-05-22
Manual pull of the 2024 data

juandlt-csu:
Hey Sam, here is the code that I used to do the manual pull via the fcw.qaqc package. Most of it is the same as before, the biggest difference being the set up for the sites in the network check, so a custom network check was made in the script. 

- The actual data should be uploaded to `data/hydrovu_2024_data/flagged/`. 
- The site list that I used is in the script in the `sites` object. 
- The date times will be in UTC.

Let me know if any thing looks off. Doing this has made me realize that the pwqn and fcw.qaqc repos have diverged a lot and that I should probably reconcile this divergence ASAP.

### DIFF
--- a/scratch/api_pull/pull_2024.R
+++ b/scratch/api_pull/pull_2024.R
@@ -1,0 +1,438 @@
+# loading packages
+package_loader <- function(x) {
+  if (x %in% installed.packages()) {
+    suppressMessages({
+      library(x, character.only = TRUE)
+    })
+  } else {
+    suppressMessages({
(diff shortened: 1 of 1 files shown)


SamStruthers:
Thanks for pulling this
It appears that FDOM was not getting pulled for any sites. I checked the raw parquet files and it isn't there either...
From reading `api_puller.R`, it doesn't appear that any parameters are deselected so I'm a bit lost as to why it is getting dropped

PR #170, opened 2025-07-31
Implement Sensor Calibration Workflow Functions

juandlt-csu:
Hey Katie, sorry for this behemoth PR, hopefully it's easy to follow along.

Sam I am adding you for the exposure, I am not expecting a full review from you :).

AI disclosure: I did use AI to generate this PR description, and then I edited it

----
This PR implements the back-calibration system for water quality sensor data from In-Situ Aqua TROLL sondes. The system corrects sensor readings by applying slope/offset corrections from HTML calibration files and temporal interpolation between calibrations to generate the calibrated time series data.

**Note:** While drift correction functions (very much inspired by the driftR package) are implemented and remain in the workflow, I am using the linear interpolation corrections as the primary calibration method.

### New Core Functions (`src/`)
- `load_calibration_data.R` - Main data loading interface
- `cal_extract_markup_data.R` - HTML parsing and extraction (_heavily_ modified)
    - This function holds the calibration extraction stuff from the previous PR and incorporates the bench top integrations. Important stuff, please pay special attention to this in the review, since if this is not good, then the downstream stuff is not good. 
- `cal_join_sensor_calibration_data.R` - Temporal data joining 
- `cal_prepare_sensor_calibration_data.R` - Calibration window creation
- `cal_back_calibrate.R` - Main orchestrator function

### New Calibration Processing Functions (`src/`)
- `cal_wt.R` - Temporal weight calculation
- `cal_inv_lm.R` - Inverse linear model transformation
- `cal_lin_trans_lm.R` - Linear transformation with temporal weighting
- `cal_lm_pH.R` - pH-specific linear model (pH → mV conversion)
- `cal_lin_trans_inv_lm_pH.R` - pH-specific inverse transformation (mV → pH)
- `cal_check.R` - Validation and final value selection

### New Drift Correction Functions (`src/`)
- `cal_one_point_drift.R` - Single-point drift correction (most sensors)
- `cal_two_point_drift.R` - Two-point drift correction (turbidity)
- `cal_three_point_drift_pH.R` - Three-point drift correction (pH)

### 1. **Check the data flow**
Follow this sequence to understand data processing:
```
load_calibration_data() → cal_join_sensor_calibration_data() → 
cal_prepare_sensor_calibration_data() → cal_back_calibrate()
```

### 2. **Review the orchestrator function to  understand the core calibration sequence**
Review `cal_back_calibrate.R` to understand the three calibration pathways:
- Standard sensors (Simple linear transformation + 1-point drift)
- Turbidity sensors (Simple linear transformation + 2-point drift)
- pH sensors (Multi-segment calibration with mV↔pH conversions + 3-point multi-segment drift)

For standard sensors, review these functions in order:
```
cal_wt() → cal_inv_lm() → cal_lin_trans_lm() → 
cal_one_point_drift() → cal_check()
```

##### **Temporal Interpolation**
- `cal_wt()` creates weights (0→1) across calibration windows
- `cal_lin_trans_lm()` applies weighted interpolation between calibration parameters
- This enables smooth transitions between calibrations over time

Turbidity is very similar, but with a 2-point drift calibration, but since we are not considering the drift corrections right now, that one should be a quick review. 

### 3. **pH-specific workflow**
pH sensors require special attention due to unit conversions:
```
cal_wt() → cal_lm_pH() → cal_lin_trans_inv_lm_pH() → 
cal_three_point_drift_pH() → cal_check()
```
The slope and offset information requires a unit conversion and then a multi-segment linear transition. Weird stuff, and I am not sure that this was done correctly since the Nernst equation that is used by the sensors to get the pH reading requires temperature to get the value. I am not sure that I can just convert these values to their "raw" readings in the same way as the other parameters without temperature information. But assuming that I don't have to do that I was able to do a linear transformation on the data, similar to the linear transitions we do for the other parameters. The pH stuff is the shakiest component in the PR, so special attention there would be appreciated. ~Though, I would not be surprised if those functions just don't work in their current state.~ The pH workflow should run without errors now, but that does not mean that the logic is perfectly sound, so a review on that is still needed. 

### 4. **Review the outputs from the examples in the workflow:**
~I did not explore many more options, though I am sure that there are other options that you can test. I did not test any pH yet. But for these examples check that the outputs and logic of the workflow make sense.~ 

You should be able to review all the outputs from the calibration workflow. There is still some stuff that needs to get fixed (Spc NA offsets → 0, checking if the structure for inputs is _more_ correct, joining on site visits, etc.), but this workflow is in a stable spot now in terms of execution and syntactical errors, but it's logic needs to be reviewed still.

### 5. **Verify function documentation:**
AI disclosure: I used AI to write the documentation for the functions and the comments within the functions. I did read them over to verify their accuracy, but if you spot anything that looks suspicious please flag it. 

**Note on error handling**
- ~I will need to add more robust error handling to handle whatever edge cases may arise as I continue to test. This is why the back calibration step breaks on the older data, haven't handled those edge cases.~ 
- The execution errors should be handled for the data that we have. The error handling is mainly around the calibration function inputs, so there is a lot that is not accounted for still but the script runs now. 

This PR is big so if you would like to talk about it or review any part of it IRL let me know. Thanks, Katie!

### DIFF
--- a/scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
+++ b/scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
@@ -245,7 +245,7 @@ while maintaining the overall automated workflow for standard cases.
 
 ## The `back_calibrate()` Function
 
-The `back_calibrate()` function serves as the orchestrator for the entire calibration 
+The `cal_back_calibrate()` function serves as the orchestrator for the entire calibration 
 process within each chunk. It automatically determines the appropriate calibration 
 workflow based on the sensor parameter type. Standard sensors (Chlorophyll-a, 
 FDOM, ORP, Pressure, Specific Conductivity, RDO) follow a five-step process 
--- a/scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
+++ b/scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
@@ -3,10 +3,16 @@ title: "fixing_calibrations"
 output: html_document
 ---
 
-```{r}
-# Initialize Environment
-# source(file.path("scratch","data_tidying","sensor_calibration","R","init_sub_env","init.R"))
+# Introduction
 
+This vignette demonstrates the complete workflow for back-calibrating water 
+quality sensor data using calibration information extracted from HTML calibration 
+files. The process corrects sensor drift and applies temporal interpolation 
+between calibrations to produce accurate time series data.
+
+# Prepare Environment
+
+```{r}
 # Load necessary packages:
 source("src/package_loader.R")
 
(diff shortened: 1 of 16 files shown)


scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,329 @@ CRAN_packages <- c(
...
+calibration approaches. The exposed nested maps allow users to easily extract 
+specific years, site-parameters, or individual chunks for manual processing 
+while maintaining the overall automated workflow for standard cases.
+
+## The `back_calibrate()` Function
+
+The `back_calibrate()` function serves as the orchestrator for the entire calibration 
+process within each chunk. It automatically determines the appropriate calibration 
+workflow based on the sensor parameter type. Standard sensors (Chlorophyll-a, 
+FDOM, ORP, Pressure, Specific Conductivity, RDO) follow a five-step process
```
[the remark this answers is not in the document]
juandlt-csu: If we decide to fully incorporate and use the drift correction functions yes, otherwise no.

juandlt-csu:
Hey @SamStruthers - thanks for bringing up the sensor swap calibration issue. After chatting with @kathryn-willi, I realize you both were definitely right about joining on the sensor serial number. I was getting hung up on the idea that sensor swaps "break" our data continuity, and initially thought that made back-calibration impossible. But, actually, I was Just Wrong lol. You're both right that it’s possible to solve this problem via the sensors’ SNs. I was super hung up on trying to force smooth transitions between different sensors (which creates those fake relationships), but actually we don’t have to force those transitions and we will just embrace segmented calibrations where each sensor gets its own calibration story. Took me a while, but I got there too haha.

I haven’t been able to incorporate the idea, and it’s definitely going to be a lift, but here is my understanding of how we deal with the sensor swap issue generally:
* Sensor A gets calibrated(1) → deployed → retrieved and calibrated(2) again. We use only A's calibrations (1 and 2) for A's data.
* Sensor B gets the same treatment with B's calibrations for B's data.
* No more trying to blend calibrations between different physical sensors, and we will just have segments instead. The “gaps” that we should see here are what were causing me concern, but actually that’s just to be expected with sensor swaps.

We're already joining by sensor serial number, so the main work is figuring out the temporal sequence of calibrations to add this additional information. Then, we will need to tweak our back-calibration functions to handle this updated framework. Not a super light lift, but a lot of the underlying structure is actually fine here.

I was definitely not seeing the vision here, but you both had the right intuition from the start. Thanks for bearing with me while I took my time to got there too, haha. That being said, as the workflow is right now we are definitely mixing sensor swap calibrations, so that will need to get fixed.

src/cal_join_sensor_calibration_data.R
```diff
@@ -0,0 +1,61 @@
...
+
+          # Extract data for current site-parameter combination
+          sensor_data <- year_sensor_data[[site_param]]
+          calibration_data <- select(year_calibration_data[[site_param]], -sensor)
+
+          # Join calibration data to sensor data using temporal proximity matching
+          # For each sensor timestamp, find the most recent calibration that
+          # precedes or equals the sensor measurement time
+          by <- join_by(site, closest(DT_round >= file_date))
+          joined_sensor_calibration_data <- left_join(sensor_data, calibration_data, by) %>%
```
kathryn-willi, line 49: so, this allows the cals to be joined regardless of whether there was a listed site visit. 🆒

scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,288 @@ CRAN_packages <- c(
...
+The `prepare_sensor_calibration_data()` function segments the joined data into 
+calibration windows bounded by consecutive calibrations. Each window represents 
+a time period where linear interpolation between two calibrations can be applied. 
+This segmentation creates manageable chunks that allow for temporal weighting of 
+calibration parameters.
+
+# Back Calibration
+
+```{r Apply back calibration functions to the data}
+# NOTE: THIS WILL NOT RUN YET!
```
kathryn-willi, line 218:
```suggestion

```

It ran for me without error!

scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,288 @@ CRAN_packages <- c(
...
+
+The explicit nested structure in  the `back_calibration()` function call serves 
+a purpose for future functionality. While automated calibration works well for 
+most situations, some sensor data may require manual intervention or custom 
+calibration approaches. The exposed nested maps allow users to easily extract 
+specific years, site-parameters, or individual chunks for manual processing 
+while maintaining the overall automated workflow for standard cases.
 
-write_rds(subset_split, here("data", "calibration_reports", "0_cal_data_munge", "munged_calibration_data.RDS"))
+## The `back_calibrate()` Function
```
kathryn-willi, line 247:
```suggestion
## The `cal_back_calibrate()` Function
```

scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,288 @@ CRAN_packages <- c(
...
+
+The `back_calibrate()` function serves as the orchestrator for the entire calibration
```
kathryn-willi, line 249:
```suggestion
The `cal_back_calibrate()` function serves as the orchestrator for the entire calibration 
```

scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,288 @@ CRAN_packages <- c(
...
+## Understanding the Nested Map Structure
+
+The back calibration process uses an explicit nested map approach that differs 
+from the simpler function calls used in previous processing steps. This design 
+choice provides three levels of iteration: years, site-parameters within each 
+year, and calibration chunks within each site-parameter. While `join_sensor_calibration_data()` 
+and `prepare_sensor_calibration_data()` handle similar nested year/site-parameter 
+iterations internally, they abstract away this complexity from the user.
+
+The explicit nested structure in  the `back_calibration()` function call serves
```
kathryn-willi, line 240:
```suggestion
The explicit nested structure in  the `cal_back_calibration()` function call serves 
```

src/cal_back_calibrate.R
```diff
@@ -0,0 +1,84 @@
...
+  if (parameter == "Turbidity") {
+    back_calibrated_chunk <- prepped_snsr_cal_df %>%
+      # Calculate temporal weights for calibration interpolation
+      cal_wt(df = ., dt_col = "DT_round") %>%
+      # Convert observed values back to raw instrumental units
+      cal_inv_lm(df = ., obs_col = "mean", lm_coefs_col = "calibration_coefs") %>%
+      # Apply temporally-weighted linear transformation between calibrations
+      cal_lin_trans_lm(df = ., raw_col = "mean_raw", lm_coefs_col = "calibration_coefs", wt_col = "wt") %>%
+      # Apply two-point drift correction for turbidity-specific drift patterns
+      cal_two_point_drift(df = ., lm_trans_col = "mean_lm_trans", drift_corr_col = "drift_input", wt_col = "wt") %>%
```
kathryn-willi, line 55: I now understand why the `mean_lm_trans` values look so different than the `mean` column. It basically is doubling down on the same information... since the slope and offset are created based on the pre-/post- sensor values

src/cal_check.R
```diff
@@ -0,0 +1,39 @@
...
+#' @seealso [cal_two_point_drift()]
+#' @seealso [cal_three_point_drift_pH()]
+
+cal_check <- function(df, obs_col, lm_trans_col) {
+
+  # Create output column name for final calibrated values
+  transformed_col <- paste0(obs_col, "_cal")
+
+  # Check if all linearly transformed values are missing (calibration failure)
+  all_na <- all(is.na(df[[lm_trans_col]]))
```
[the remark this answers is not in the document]
kathryn-willi: I can't find any instances in the datasets where this happens. Could you provide an example of when this occurs?

kathryn-willi (approved):
This is excellent - great work!!! As we have discussed, we will want to implement the back calibration steps but not the drift steps, but I'm happy they were provided here so that I can try to use them as a starting point for sensor drift. 

Based on how this workflow was designed I believe it can be woven into any location in our auto-QA/QC pipeline after we "tidy" the data (namely aggregating the raw values into the 15-minute `mean`) - is that correct? This built-in flexibility is great.

Now that this PR is approved, I suggest adding these functions into the {fcw.qaqc} package, and adding the fixing_calibrations.Rmd as a vignette. I also suggest incorporating the sensor swap stuff to {fcw.qaqc} directly. Once again, great job!

src/cal_back_calibrate.R
```diff
@@ -0,0 +1,84 @@
...
+#' @seealso [cal_wt()]
+#' @seealso [cal_inv_lm()]
+#' @seealso [cal_lin_trans_lm()]
+#' @seealso [cal_one_point_drift()]
+#' @seealso [cal_two_point_drift()]
+#' @seealso [cal_lm_pH()]
+#' @seealso [cal_lin_trans_inv_lm_pH()]
+#' @seealso [cal_three_point_drift_pH()]
+#' @seealso [cal_check()]
+
```
SamStruthers, line 23: explaining the differences in the columns returned between parameters would be helpful here. I was a bit confused trying to figure out which column was which between turbidity vs pH

SamStruthers (approved):
Everything ran on my end! If you could provide a quick explanation of the different columns returned for different parameters, that would be super helpful for folks picking this up downstream

src/cal_extract_markup_data.R
```diff
@@ -1,70 +1,248 @@
...
+      ) %>%
+        # Standardize site names using fix_sites() function
+        fix_sites() %>%
+        # Standardize sensor parameter names to match sensor data conventions
+        dplyr::mutate(
+          sensor = dplyr::case_when(
+            sensor == "chlorophyll_a" ~ "Chl-a Fluorescence",
+            sensor == "conductivity" ~ "Specific Conductivity",
+            sensor == "fdom" ~ "FDOM Fluorescence",
+            sensor == "p_h" ~ "pH",
```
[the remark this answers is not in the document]
juandlt-csu: This is just what happens to pH when `janitor()` functions are applied to it to clean it up.

scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,288 @@ CRAN_packages <- c(
...
+  
+  sensor_data <- names(calibration_data) %>% 
+    map(function(year){
+      
+      # Grab the site for the year from the calibration_data[[year]]
+      site_list <- map(names(calibration_data[[year]]), ~word(.x, 1, sep = "-")) %>% 
+        unlist() %>% 
+        unique(.) 
+      
+      site_list <- unique(c("tamasag", "legacy", "lincoln", "timberline", "prospect", "boxelder", "archery", "river bluffs", site_list))
```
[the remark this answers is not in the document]
juandlt-csu: This was my workaround for dealing with the different names that the sites have had over the years. My thinking here was that the most recent data has different names for the the same sites that we had before, and eventually we want to join all of it by site (and parameter) so doing it this way let me pull everything and then fix the names later.

scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,288 @@ CRAN_packages <- c(
...
+      new_data <- fcw.qaqc::munge_api_data(api_dir = temp_dir) %>% 
+        fix_sites() %>%
+        split(f = list(.$site, .$parameter), sep = "-") %>%
+        purrr::keep(~ nrow(.) > 0)
+      
+      # Tidy the API data
+      message("Starting Tidy Step")
+      sites <- unique(dplyr::bind_rows(new_data) %>% dplyr::pull(site))
+      
+      params <- c("Chl-a Fluorescence", "Depth", "DO", "ORP", "pH", "Specific Conductivity",
```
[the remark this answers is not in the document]
juandlt-csu: Too true, I will have to add that back in here.

scratch/data_tidying/sensor_calibration/fixing_calibrations.Rmd
```diff
@@ -27,67 +33,288 @@ CRAN_packages <- c(
...
+Each site-parameter entry includes calibration coefficients (slope and offset values) 
+and drift correction measurements (pre- and post-calibration standard readings),
+along with their related identification and datetime information.
+
+# Load Sensor Data
+
+```{r load-snsr-data}
+# Read in 2022, 2023, 2024, and 2025 sensor data
+if (file.exists(here("data", "manual_data_verification", "complete_dataset", "23_25_hv_pull"))) {
+  sensor_data <- read_rds(here("data", "manual_data_verification", "complete_dataset", "23_25_hv_pull")) %>%
```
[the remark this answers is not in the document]
juandlt-csu: Yep, as it is this is raw post-munged pre-autoqaqc hydrovu data since I couldn't find where all the "actual" data was. I will add the site visit flags to help with the calibration joins, though.

src/cal_join_sensor_calibration_data.R
juandlt-csu: Yes, but in the future I think there will be some connection to the latest site visit, and it won't just be joined straight up like this to the data. I think that that should also make the sensor swap issues more manageable.

juandlt-csu:
I am going to implement the changes requested in this PR in the transfer to `{fcw.qaqc}`. For now I am going to close this PR, and once the final changes are implemented to `{fcw.qaqc}` I will close this branch.

PR #179, opened 2025-09-29
2019-2021 data processing

kathryn-willi:
This PR contains a .RMD that passes the old (2019-2021) time series data though our auto-QAQC processes. This .RMD is largely based off of the `2025_Q2_pull.R` script in `scratch/api_pull`. As a reminder, this data came from a time where our field and data collection processes were very different (e.g., no field notes; data not always on HydroVu, etc.).

I ran the pipeline myself, and it seems to be doing what we want it to. The stored outputs are identical in structure to our modern data products. However! The one thing I'd love additional eyes on is the time stamp. I believe all are in UTC, but if you could take a look at some diurnal patterns to confirm, that would be amazing.

Lastly, I went through the repo and deleted any functions and scripts that are now contained in our {fcw.qaqc} to avoid future confusion.


SamStruthers:
Thanks for organizing this Katie! 
Here is a plot of Depth at ELC (flow at USGS gage), both in UTC:
Screenshot 2025-09-29 at 11 57 13 AM
The storm events seem to coincide correctly for 2021

This does not appear to be the case for 2020 ( which appears to be off by 7 hours):
Screenshot 2025-09-29 at 12 02 20 PM


It looks like the `timestamp` column is MST or MT for 2019 & 2020 years but not 2021...
Here are some plots of median values by hour to justify.

As is for 2021 (assuming "correctly" timestamp in historical data is UTC):
image
As is with 2020 (data "incorrectly in UTC"):
image

Correcting 2020 (from mst -> utc):
image

Raw & Corrected for 2019 and 2025 for ref: 
image

PR #180, opened 2025-10-07
DT fixes part 694830485435

kathryn-willi:
What I think truly truly this time is the fix of the DT timezone issues! Sam was able to deduce an issue that was occurring where timestamps were mislabelled as UTC by HydroVu but were actually in the timezone of the computer that downloaded them.

### DIFF
--- a/data_organization/Process_2019_2021.rmd
+++ b/data_organization/Process_2019_2021.rmd
@@ -109,7 +109,10 @@ vulink_reader <- function(file) {
     slice(-1:-31) %>%
     janitor::row_to_names(row_number = 1) 
 }
+```
 
+
+```{r}
 ## Rist
 
 hydrovu_utc <- map_dfr(list.files("data/sensor_data/2018_2021/site_organized_hydrovu/rist/", full.names = T), hydrovu_reader) %>%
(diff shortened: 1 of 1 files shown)


data_organization/Process_2019_2021.rmd
```diff
@@ -134,10 +137,12 @@ hydrovu_utc <- map_dfr(list.files("data/sensor_data/2018_2021/site_organized_hyd
   mutate(DT = as_datetime(DT, tz = "UTC")) 
 
 rist <- hydrovu_utc %>%
-  # mutate(DT = round_date(DT, "15 minutes")) %>%
   mutate(year = lubridate::year(DT)) %>%
   mutate(date = as_date(substr(DT, 1, 10))) %>%
   distinct(.keep_all = TRUE) %>%
+  # Force time zones appropriately. MST to UTC:
+  mutate(DT = case_when(year %in% c(2019, 2020) ~ DT + hours(7),
```
SamStruthers, line 144: How did you double check this was in MST and not MT? It looks like other sites you do other conversions

SamStruthers (approved):
Looking at the data (using the same script I gave you), things appear to line up correctly with our 2025 UTC data! If you could just provide some info/comments on how you determined which site/years were in MST vs MT vs MDT that would be great for future reference! Otherwise it looks good and excited to check this off our list!!!

image

image

image

